PR #221, opened 2026-02-09
Respect theme spacing presets instead of always overriding

jnealey-godaddy:
## Description
This PR changes the spacing preset logic to respect theme-defined `spacingSizes` instead of always overriding them with standard WordPress spacing slugs. Previously, the code would always add WP standard spacing slugs (20, 30, 40, etc.) even when themes defined their own spacing system, which could cause conflicts with theme-specific spacing implementations (e.g., TT5's responsive `clamp()` values being replaced with hardcoded rem).

## Type of Change
- [x] Code refactoring

### Spacing presets are now fallbacks only
- Spacing sizes (standard WP slugs + legacy DSG slugs) are only added when the theme doesn't define its own `spacingSizes`
- `spacingScale` configuration moved inside the conditional block

### Block spacing inherits from theme
- Container, Tabs, and Tab block spacing (padding, margin, blockGap) is no longer unconditionally injected
- When a theme defines spacing, blocks inherit the theme's spacing rhythm via `blockGap` support and the theme's preset variables
- Block spacing defaults are only applied when DSG controls the spacing scale
- Non-spacing block styles (border, color, typography, elements) are always applied as defaults

### CSS fallbacks for robustness
- Spacing values now include CSS `var()` fallbacks (e.g., `var(--wp--preset--spacing--40, 1.5rem)`) for edge cases where preset variables don't exist

## Why this approach is correct
- Standard WP spacing slugs (20-80) are an established convention — TT5, TT4, and WP core defaults all define these slugs
- `update_with()` at the `wp_theme_json_data_theme` filter replaces the entire `spacingSizes` array, not slug-level merging — always adding DSG presets would overwrite theme values
- Blocks referencing `var(--wp--preset--spacing--40)` resolve against whatever the active theme provides, which is the intended behavior

## Testing
- [ ] Tested on frontend
- [ ] Tested with Twenty Twenty-Five theme (has `spacingSizes` with `clamp()` — DSG should NOT override)
- [ ] Tested with theme that doesn't define `spacingSizes` (DSG fallback spacing should apply)
- [ ] Verified Container/Tabs/Tab blocks inherit theme spacing when theme has `spacingSizes`
- [ ] Verified Container/Tabs/Tab blocks get DSG spacing defaults when theme lacks `spacingSizes`
- [ ] No console errors
- [ ] No PHP errors

## Checklist
- [x] I have updated documentation/comments as needed
- [x] PHP linting passes
- [x] Build passes

### DIFF
--- a/includes/admin/class-global-styles.php
+++ b/includes/admin/class-global-styles.php
@@ -80,16 +80,44 @@ class Global_Styles {
 			),
 		);
 
-		// Only add spacing presets when the theme doesn't define its own.
-		// Themes that define spacingSizes (e.g. with clamp() or px values) should
-		// not be overwritten — blocks using var(--wp--preset--spacing--XX) will
-		// resolve against whatever the active theme provides.
+		// Only add spacing presets and block spacing defaults when the theme
+		// doesn't define its own spacingSizes. Themes that define spacingSizes
+		// (e.g. with clamp() or responsive values) should not be overwritten.
+		// Blocks inherit the theme's spacing rhythm via blockGap support and
+		// the theme's spacing presets resolve automatically.
 		if ( ! $this->theme_has_spacing_sizes( $theme_data ) ) {
 			$dsg_settings['settings']['spacing']['spacingSizes'] = array_merge(
 				$this->get_standard_spacing_sizes(),
 				$this->get_legacy_spacing_sizes()
 			);
 			$dsg_settings['settings']['spacing']['spacingScale'] = array( 'steps' => 0 );
+
+			// Block spacing defaults — only applied when DSG controls the spacing
+			// scale. When theme has spacing, blocks inherit from the theme's system.
+			$dsg_settings['styles']['blocks']['designsetgo/container']['spacing'] = array(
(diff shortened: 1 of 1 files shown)


jnealey88:
The standard WP spacing slugs (20, 30, 40, 50, 60, 70, 80) are a widely-adopted convention. TT5, TT4, and WordPress core defaults all define these slugs. When themes define their own spacingSizes, the variables --wp--preset--spacing--40 etc. resolve against the theme's values — that's the intended behavior.

Option A would reintroduce the bug this PR fixes. update_with() within the wp_theme_json_data_theme filter replaces the entire spacingSizes array, not slug-level merging. Always adding DSG's presets would overwrite TT5's responsive clamp() values with hardcoded rem values.

The edge case of a theme using entirely non-standard slugs (not 20-80) is real but extremely rare and goes against WordPress spacing conventions. If we need to address it, the right fix is CSS fallback values in the block stylesheets (e.g., var(--wp--preset--spacing--40, 1.5rem)), not overwriting theme spacing.

PR #342, opened 2026-04-12
feat: add admin-ajax.php fallback for rate-limited REST API

jnealey-godaddy:
## What Changed
This PR hardens the form submission fallback paths for hosts that rate-limit or otherwise interfere with the REST API.

It now:
- falls back from the REST endpoint to `admin-ajax.php` when the REST request is rate-limited
- supports non-AJAX form submission through `admin-post.php`
- preserves Turnstile verification across both AJAX and non-AJAX submission paths
- scopes redirect-based success and error messaging to the form that was actually submitted
- aligns generated phone field metadata with the runtime validator by using `tel`
- adds PHPUnit coverage for server-side field type resolution and client type spoofing protection

## Why
Some environments were blocking or rate-limiting `/wp-json/`, which made otherwise valid forms fail. The fallback path fixed the transport problem, but it still had edge cases:
- non-AJAX submissions were not carrying the Turnstile token
- redirect-based messaging could affect every form on the page instead of the submitted one
- phone field metadata was inconsistent between generated HTML and the main form runtime
- the server needed explicit test coverage proving that saved form definitions win over client-submitted field types

## Impact
Users get a more reliable form experience on restrictive hosts, and form validation/security behavior is now more consistent across transport modes.

For developers:
- non-AJAX submissions now pass the Turnstile token through the standard handler
- form status redirects include the form id so multi-form pages behave correctly
- server-side validation continues to rely on the saved form definition rather than trusting client field types

## Validation
I validated this with:
- `npm run lint:js`
- `composer run-script lint`
- `npm run build`

Full PHPUnit execution is still blocked locally because the WordPress test suite is only configured through `wp-env`, and `wp-env` could not run in this shell because Docker is not running.

### DIFF
--- a/includes/abilities/class-form-field-html-generator.php
+++ b/includes/abilities/class-form-field-html-generator.php
@@ -458,7 +458,7 @@ class Form_Field_Html_Generator {
 		if ( $help_text ) {
 			$html .= ' aria-describedby="' . esc_attr( $field_id ) . '-help"';
 		}
-		$html .= ' data-field-type="phone"/>';
+		$html .= ' data-field-type="tel"/>';
 
 		$html .= self::generate_help_text( $field_id, $help_text );
 		$html .= '</div>';
--- a/includes/blocks/class-form-handler.php
+++ b/includes/blocks/class-form-handler.php
@@ -405,11 +405,11 @@ class Form_Handler {
 			);
 		}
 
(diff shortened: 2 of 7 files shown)


jnealey-godaddy:
@claude review the latest commits

PR #345, opened 2026-04-14
fix(forms): apply block.json defaults when reading form attrs server-side

jnealey-godaddy:
## Summary

Contact form submissions on production silently stopped sending admin-notification emails. Every submission recorded `_dsg_email_sent` as missing (`—` in the admin list), meaning `wp_mail()` was never called.

**Root cause:** `parse_blocks()` only returns attributes that were serialized into the block comment, and the editor omits any attribute equal to its declared default. `enableEmail` defaults to `true` in [src/blocks/form-builder/block.json](../blob/main/src/blocks/form-builder/block.json#L132-L134), so the key was absent from `$block_attrs` for any form left at default settings. The guard in [class-form-handler.php:699](../blob/main/includes/blocks/class-form-handler.php#L699) — `empty( $block_attrs['enableEmail'] )` — returned early, before `wp_mail()` or any meta write. Introduced by #240 (server-side email config); client-side code previously relied on React attribute defaults, so the regression only manifested on the server.

## Fix

- Merge registered block-type defaults into parsed attributes inside `find_form_block_attributes()` via a new `apply_form_block_defaults()` helper. All three server-side consumers (rate-limit at line 249, Turnstile check, email notification at line 697) now see the same attribute set the editor does.
- Fully-qualified `\WP_Block_Type_Registry` to avoid namespace resolution in `DesignSetGo\Blocks`.
- Bumped the form-attrs transient cache prefix to `dsgo_form_attrs_v2_` so stale pre-fix caches are invalidated automatically on deploy (no manual transient flush required). Updated `invalidate_form_block_transients()` and the corresponding test references.

## Test plan

- [x] `php -l` clean on modified files
- [x] `composer run lint` passes (79/79)
- [x] Manual end-to-end verification on designsetgoblocks.com — submission #2036 recorded `_dsg_email_sent=yes` with `To: justnealey@gmail.com` and `wp_mail()` returned true (confirmed in admin submission detail)
- [ ] Run `tests/phpunit/form-handler-test.php` in CI (cache-key assertions updated to v2 prefix)
- [ ] Smoke check a form with `enableEmail` explicitly toggled off still skips sending (backward-compat path)

### DIFF
--- a/includes/blocks/class-form-handler.php
+++ b/includes/blocks/class-form-handler.php
@@ -940,7 +940,9 @@ class Form_Handler {
 	 */
 	private function get_form_block_attributes( $form_id ) {
 		// Check transient cache first to avoid LIKE queries on every submission.
-		$cache_key = 'dsgo_form_attrs_' . md5( $form_id );
+		// v2 prefix invalidates older caches that were stored before block-type
+		// defaults were merged into parsed attributes.
+		$cache_key = 'dsgo_form_attrs_v2_' . md5( $form_id );
 		$cached    = get_transient( $cache_key );
 
 		if ( false !== $cached ) {
@@ -996,7 +998,12 @@ class Form_Handler {
 				isset( $block['attrs']['formId'] ) &&
 				$block['attrs']['formId'] === $form_id
 			) {
-				return $block['attrs'];
+				// parse_blocks() returns only the attributes that were serialized into
+				// the block comment. The editor omits attributes that equal their
+				// declared default, so booleans like `enableEmail` (default true) and
+				// similar may be missing here. Merge in the block-type defaults so
+				// server-side consumers see the same attribute set the editor does.
+				return $this->apply_form_block_defaults( $block['attrs'] );
 			}
 
 			if ( ! empty( $block['innerBlocks'] ) ) {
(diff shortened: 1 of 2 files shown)


jnealey-godaddy:
Addressed review feedback in 4a3b77a5:

- **High (test_client_email_params_are_ignored will break):** Verified against source — fixture at line 528 had `{"formId":"noemail1"}` with no `enableEmail`. After the fix, `apply_form_block_defaults()` would merge in `enableEmail:true` and `assertEmpty($email_sent)` would fail. Fixture now explicitly sets `"enableEmail":false` so the test exercises the intended "email disabled" path hermetically.
- **Medium (no test for default injection):** Added `test_get_form_block_attributes_merges_block_json_defaults` with a bare `{"formId":"..."}` block asserting that `enableEmail`, `emailSubject`, `enableHoneypot`, and `enableTurnstile` all get filled in — covers both truthy and `false` defaults, and acts as a direct regression guard for the "emails never send" bug.
- **Low (orphaned v1 transients):** Leaving as-is — v1 keys auto-expire in ≤1h and pruning them would require a one-shot migration on activation that isn't worth the complexity for short-lived cache.

Issue #12, opened 2025-11-15
Author Box Block


jnealey88:
## Block: Author Box Block

**Version:** v1.6.0  
**Priority:** Low | **Effort:** Low | **Impact:** Low

### Problem
Blog posts need author bio sections for credibility and social links.

### Features

**Content:**
- Author avatar (Gravatar or custom upload)
- Author name
- Bio/description
- Social media links (Twitter, LinkedIn, Website, etc.)
- Email link (optional)
- Author archive link ("View all posts by...")

**Layout:**
- Horizontal (image left, content right)
- Vertical (image top, content below)
- Minimal (no image, just text)

**Styling:**
- Background color/border
- Avatar shape (circle, square, rounded)
- Avatar size
- Social icon styles
- Padding/spacing

**Features:**
- Auto-pull author data (from WordPress user)
- Manual override (custom author for guest posts)
- Multiple authors support

### Use Cases
- Blog post author bio
- Guest post credits
- Team member profiles
- Podcast host info

### Jobs to be Done
"When I publish articles, I want author bios with social links, so readers connect with writers and follow them."

Issue #11, opened 2025-11-15
Related Posts Block


jnealey88:
## Block: Related Posts Block

**Version:** v1.6.0  
**Priority:** Medium | **Effort:** Medium | **Impact:** Medium

### Problem
Keep users on site longer by suggesting related content.

### Features

**Relation Logic:**
- By category (same categories)
- By tag (shared tags)
- By custom taxonomy
- By author (same author)
- Manual selection (choose specific posts)
- Random

**Settings:**
- Number of posts (2-12)
- Exclude current post
- Order by (date, random, relevance)
- Heading ("Related Posts", "You Might Like", etc.)

**Layout:**
- Grid (2-4 columns)
- List
- Carousel/Slider

**Display:**
- Featured image
- Title
- Excerpt (short)
- Date/author
- Category

**Styling:**
- Card backgrounds
- Hover effects
- Image aspect ratio
- Spacing

### Use Cases
- Blog post endings
- Article suggestions
- Portfolio related work
- Product cross-sell

### Jobs to be Done
"When readers finish an article, I want to suggest related content, so they stay engaged and explore more of my site."

Issue #10, opened 2025-11-15
Table of Contents Block


jnealey88:
## Problem
Long-form content needs navigation. Users manually create anchor links or install heavy plugins.

## Proposed Solution
Auto-generated table of contents with smooth scrolling, scroll spy, and sticky positioning option.

## Key Features
- Auto-scans page for headings (H2-H6)
- Generates anchor links automatically
- Updates when headings change
- Selectable heading levels
- Hierarchical (nested) or flat list
- Ordered or unordered list
- Custom TOC heading
- Smooth scroll to section
- Scroll spy (highlight current section)
- Collapsible on mobile
- Sticky TOC option
- Keyboard navigation
- Screen reader friendly

## Use Cases
- Long blog posts
- Documentation pages
- Tutorial content
- Legal pages
- Guides and resources

## Jobs to be Done
"When I write long articles, I want an auto-updating table of contents, so readers navigate easily and stay engaged."

## Effort Estimate
Medium

## Impact
High

---
**Category:** Dynamic Content
**Source:** BLOCKS-ROADMAP.md
**Version:** v1.6.0

PR #372, opened 2026-04-20
feat(query): Dynamic Query v2.3 — nested loops, relationships, visibility, group-by

jnealey-godaddy:
## Summary

v2.3 closes the remaining headline gap with Bricks / Elementor Pro / JetEngine by adding four features to the Dynamic Query family:

- **Relationship source** reads a parent-context field (meta or ACF relationship) and iterates referenced posts via `post__in`; fallback modes for when no IDs resolve.
- **Nested loops with parent context** — outer Query's current item flows into inner Queries via a new `designsetgo/parentItem` block context. DSGo bindings (`designsetgo/post-meta`, `designsetgo/acf`) now accept a `scope` arg (`'self'` default / `'parent'` / `'root'`) that reads from the parent-stack maintained by `designsetgo_query_render_item()`.
- **Conditional inner-block visibility** — every block now carries a `dsgoVisibility` attribute with an inspector panel under Advanced → Visibility. Rule types: meta / taxonomy / index / auth, combined with AND/OR relations. Server evaluator (`DesignSetGo\BlockVisibility::matches()`) and JS mirror (`src/extensions/visibility/evaluateRules.js`) stay in sync so editor previews match frontend output.
- **Group-by partitioning** — new `groupBy` attribute on the Query block partitions items by taxonomy / meta / date (with year / year-month / year-month-day precision). New sibling block `designsetgo/query-group-header` renders once per group inside a `<section class=\"dsgo-query-group\">` wrapper.

No breaking changes — all new attributes default to `null` / `'self'`. Plugin version stays at 2.1.0 since v2.2 was also unreleased; v2.3 entries land under `[Unreleased]` in the CHANGELOG.

Plan lives at [\`docs/plans/2026-04-19-dynamic-query-v2.3-nested.md\`](docs/plans/2026-04-19-dynamic-query-v2.3-nested.md).

## Test plan
- [ ] Jest unit tests green (1952/1952 on last local run)
- [ ] PHPUnit integration tests green (RelationshipRenderTest, BlockVisibilityTest, VisibilityIntegrationTest, ParentStackTest, BindingsScopeTest, GroupByTest)
- [ ] `npm run build`, `npm run lint:js`, `npm run lint:css` clean
- [ ] Manual: relationship source renders referenced posts on frontend + editor preview (matches `related_posts` meta/ACF)
- [ ] Manual: set a `dsgoVisibility` rule (meta = \"featured\" equals \"1\") on a paragraph inside a query item — verify it's hidden on non-featured posts in both the editor preview and frontend
- [ ] Manual: outer Query iterates posts; inside the item template add an inner Query; a paragraph bound with `scope: 'parent'` resolves to the outer post's meta
- [ ] Manual: Query → Group by = Taxonomy, Group taxonomy = category → frontend shows one `<section class=\"dsgo-query-group\">` per category with a Query Group Header above each group
- [ ] Regression: infinite scroll + filters + load-more + editor live preview still work inside both a flat and a grouped query

Follow-up: v2.4 (JetEngine / Meta Box / Pods bindings, PHP escape-hatch UI, JSON export/import).

### DIFF
--- a/.claude/claude.md
+++ b/.claude/claude.md
@@ -65,6 +65,16 @@ For sibling blocks that already exist and meet the "1–3 attribute difference +
 - Editor live preview: Posts via `useEntityRecords`, users/terms via `/designsetgo/v1/query/preview` REST route. First item wraps editable `InnerBlocks`; items 2..N render `BlockPreview`.
 - Admin dashboard: Settings → DesignSetGo → Dynamic Query (requires `manage_options`).
 
+### Query block family (Dynamic Query v2.3)
+
+- Relationship source (`source: 'relationship'`) reads `relationshipField` from the nearest parent-stack item and iterates referenced post IDs via `post__in`.
+- `relationshipFallback`: `'empty'` | `'all'` | `'parent'` controls behavior when the field yields zero IDs.
+- DSGo bindings (`designsetgo/post-meta`, `designsetgo/acf`) accept a `scope` arg: `'self'` (default), `'parent'`, `'root'`. Reads from `$GLOBALS['designsetgo_parent_stack']` pushed by `designsetgo_query_render_item()`.
+- Every block has a `dsgoVisibility` attribute (registered via `blocks.registerBlockType` filter in `src/extensions/visibility/filters.js`). Rule shape: `{ operator: 'AND'|'OR', rules: [{ type: 'meta'|'taxonomy'|'index'|'auth', op, key?, value }] }`. Server evaluator: `DesignSetGo\BlockVisibility::matches()`. Editor mirror: `src/extensions/visibility/evaluateRules.js`.
+- New sibling block `designsetgo/query-group-header` renders once per group inside a parent Query when `groupBy` is set.
+- `groupBy` attribute on the Query block — shape `{ field: 'taxonomy'|'meta'|'date', key: string }`. Server partitions items in `designsetgo_query_partition_items()` (in `render-helpers.php`).
+- Custom visibility rule types via `designsetgo_visibility_rule` filter (`($match, $rule, $context)` → bool|null).
+
 ### Inspector IA (Theme 3)
 
 Three panels per block, in this order: **Settings** → **Style** → **Advanced**. Use `<DsgoInspectorPanel>` (the `ToolsPanel` wrapper) for all custom inspector controls; never reach for `PanelBody` directly.
--- a/.claude/docs/QUERY-BLOCK-GUIDE.md
+++ b/.claude/docs/QUERY-BLOCK-GUIDE.md
@@ -356,6 +356,216 @@ The endpoint calls `designsetgo_query_render()` — the same function used by th
 
 ---
 
+## v2.3 Recipes
+
+### Recipe — show posts linked via an ACF/Meta relationship field
+
+An outer post stores an ACF relationship field (`related_posts`) containing an array of post IDs. Configure the Query block to iterate exactly those posts:
+
+1. Query block → Settings → **Source**: Relationship.
+2. **Relationship field**: `related_posts` (the ACF or postmeta key that stores the IDs array).
+3. **Fallback when field is empty**: choose `Render no items` (or `all` / `parent` depending on your intent).
+
+The server renderer reads the nearest parent-stack item's `related_posts` meta value, passes the IDs through `post__in`, and falls back according to `relationshipFallback`.
+
+```html
+<!-- wp:designsetgo/query {
+    "queryId":"related-products",
+    "source":"relationship",
+    "relationshipField":"related_posts",
+    "relationshipFallback":"empty"
+} -->
+<ul class="wp-block-designsetgo-query">
+
+    <!-- wp:core/post-title {"level":3} /-->
+    <!-- wp:core/post-featured-image /-->
+
+</ul>
+<!-- /wp:designsetgo/query -->
+```
+
+---
+
+### Recipe — hide a "Featured" badge on non-featured posts
+
+Use the Visibility panel on any inner block to show it only when a meta condition is met. No custom PHP required.
+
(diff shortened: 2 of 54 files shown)


jnealey-godaddy:
@claude do a code review

jnealey-godaddy:
Addressed all 12 review findings. Latest HEAD: `dc6f0794`.

**Critical bugs fixed:**
- **#1** [`ca0ba17a`](../commit/ca0ba17a): RuleRow operator keys now match both evaluators (`exists`/`not_exists` → `not_empty`/`empty`, `less_than`/`greater_than` → `lt`/`gt`, dropped unimplemented `even`/`odd`).
- **#2** [`0c43ef42`](../commit/0c43ef42): Auth rule now stores a boolean on `rule.value` instead of `rule.op`, matching both evaluators. Type-switcher now resets the rule shape so stale `op`/`key` fields can't bleed across types.
- **#3** [`a68c4f0a`](../commit/a68c4f0a) + [`3d3753d8`](../commit/3d3753d8): Per-item `index` now threaded into `designsetgo_query_render_item()` for posts, users, terms renderers, plus the grouped render path (per-group counter, so "index 0" means first-in-group).
- **#4** [`f44abc34`](../commit/f44abc34): Relationship preview swapped from invalid `useEntityRecords('postType', 'any', …)` to `'post'` with a `TODO(v2.4)` for CPT support.

**Notable fixes:**
- **#5** [`0438c6ff`](../commit/0438c6ff) + [`9b64aa06`](../commit/9b64aa06): Relationship renderer falls back to `$GLOBALS['designsetgo_parent_stack']` when `$context['parent_stack']` is empty. New integration test `test_relationship_uses_globals_parent_stack_fallback` covers the production path. Also added `get_field` → `get_post_meta` fallback for relationship fields when ACF returns `false`, plus missing `render-relationship.php` CopyPlugin rule in `webpack.config.js`.
- **#6** [`dc646f34`](../commit/dc646f34): Visibility gate returns a hidden placeholder `<div class="dsgo-visibility-hidden" aria-hidden="true" style="display:none">` instead of `null`, preserving React tree stability.
- **#7** [`1c382c84`](../commit/1c382c84): Scope resolution extracted into private `resolve_scoped_post_id( $args, $block )` method shared by `get_post_meta_value()` + `get_acf_value()`.
- **#8** [`2ad1701c`](../commit/2ad1701c): Grouped preview now walks `_embedded['wp:term']` and uses term **name** (matching server), and fallback spelling normalized to US `Uncategorized` everywhere.
- **#9**: Addressed in [`3d3753d8`](../commit/3d3753d8) — relationship preview now reads the parent's post type from `getCurrentPostType()` on `core/editor`.
- **#10** [`130cd49b`](../commit/130cd49b): JS relationship-ID normalizer mirrors the PHP shape coverage (array / CSV string / single int / object with `.id`/`.ID`). 13 new unit tests.
- **#11** [`dc6f0794`](../commit/dc6f0794): Docs corrected — `designsetgo/groupKey` + `designsetgo/groupCount` references removed; only the keys server actually provides (`designsetgo/groupLabel`, `designsetgo/groupValue`) are documented.
- **#12** [`e7611301`](../commit/e7611301): Narrowed context-filter's BLOCKED set so query-family blocks (core + DSGo) no longer receive the item-context keys they'd never use.

**Verification:** Jest 1966/1966 green, webpack build clean. PHPUnit (via docker/wp-env): all new integration tests pass including the new production-path relationship tests; the two unrelated pre-existing failures (`Test_Patterns_Loader`) are untouched by this work.

Manual editor/frontend QA on `localhost:9451` still pending — flagged in the PR description.

jnealey-godaddy:
@claude do another review

jnealey-godaddy:
Pass 2 follow-up — all 6 new findings addressed. HEAD is now `139a108a`.

**Critical (2):**
- **pass2 #1** [`4175b41c`](../commit/4175b41c) — Added a `render_block` filter that gates ALL blocks (top-level + nested inside `core/group`, `core/columns`, etc.) when the parent stack is non-empty. Registration is idempotent via static method `BlockVisibility::register()`, called once from `Plugin::load_dependencies()`. New integration test `NestedVisibilityTest::test_visibility_applies_to_nested_blocks` covers the group-wrapped case. The original loop-level check in `designsetgo_query_render_item()` stays as a top-level optimization.
- **pass2 #2** [`91c67907`](../commit/91c67907) — Editor context now emits `designsetgo/isAuthenticated: true` from `buildContext()` (editor sessions are always logged-in), and `designsetgo/isAuthenticated` is in `CONTEXT_KEYS` so `context-filter.js` propagates it onto every block's `usesContext`.

**Notable (4):**
- **pass2 #3** [`c78fd71d`](../commit/c78fd71d) — `EditorPreviewList` now routes `source === 'relationship'` through the `useQueryPreview` hook (which already handles parent-post lookup + ID normalization), so the preview gets real records instead of the always-empty REST fallback. All three hooks (`usePosts`, `useRemotePreview`, `useQueryPreview`) call unconditionally at the top; non-matching sources no-op cheaply.
- **pass2 #4** [`64810aeb`](../commit/64810aeb) — Grouped render path now emits `designsetgo/groupItemIndex` as a zero-based per-group counter (alongside the existing flat `designsetgo/itemIndex`), so authors can target "first item in group" separately from "first item overall". `CONTEXT_KEYS` + `buildContext()` + docs updated.
- **pass2 #5** [`a262decd`](../commit/a262decd) — Removed the stale "NOTE: gate is inert until Task C3" comment from `filters.js`. Task C3 has been complete for several commits.
- **pass2 #6** [`139a108a`](../commit/139a108a) — Removed the dead `showValue` always-true guard in `IndexControls`.

**Verification:** Jest 1967/1967 green, webpack build clean. PHPUnit (new `NestedVisibilityTest` added) verified via docker-backed wp-env.

Manual browser QA on `localhost:9451` still pending — flagged in the PR description.

PR #379, opened 2026-04-21
feat(form-builder): persist confirmation message across page reloads

jnealey-godaddy:
## Summary

- Stores the success message in `sessionStorage` after both AJAX and non-AJAX (redirect) form submission
- On page load, restores the confirmation message from `sessionStorage` if no URL-param redirect is present
- Scoped per form ID (`dsgo_confirmed_<formId>`) so multiple forms on the same page are handled independently

## Test Plan

- [ ] Submit a form via AJAX — confirm message appears, then reload the page and verify message still shows
- [ ] Submit a form via non-AJAX POST — confirm message appears on redirect, then reload and verify it persists
- [ ] Open the same page in a new tab — verify the confirmation does NOT carry over (sessionStorage is tab-scoped)
- [ ] Submit a form with a redirect URL configured — verify no message is stored (redirect takes over)

### DIFF
--- a/src/blocks/form-builder/view.js
+++ b/src/blocks/form-builder/view.js
@@ -182,7 +182,18 @@ function initFormBuilder() {
 		}
 
 		const formId = formContainer.getAttribute('data-form-id');
-		const storageKey = `dsgo_confirmed_${formId || 'default'}`;
+		const storageKey = (() => {
+			if (formId) {
+				return `dsgo_confirmed_${formId}`;
+			}
+			if (formContainer.id) {
+				return `dsgo_confirmed_instance_${formContainer.id.replace(/[^a-zA-Z0-9_-]/g, '_')}`;
+			}
+			const idx = Array.from(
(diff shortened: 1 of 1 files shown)


jnealey-godaddy:
Addressed all review comments in e5e2ceb:

- **Medium (never cleared):** Added `clearConfirmation()` helper; stored confirmation is now removed immediately after being displayed, so it shows on the first reload after submission only.
- **Low (shared key for no-formId forms):** Replaced the `'default'` fallback with an IIFE that tries the container's `id` attribute first, then falls back to the DOM index among all `[data-form-id]` elements, so each form instance gets a unique key.
- **Stale success masking error (redirect path):** `clearConfirmation()` is now called when showing an error redirect status.
- **Stale success masking error (AJAX path):** `clearConfirmation()` is called at the start of each submit attempt, before the async fetch, so any prior stored success is cleared regardless of the outcome.

PR #383, opened 2026-04-22
style: mirror icon-button hover on core buttons

jnealey-godaddy:
## Summary
- Adds a global `:where(.wp-block-button__link)` hover (opacity 0.9 + 1px lift, matching `dsgo-icon-button`) so core buttons share the same micro-interaction as DSGo icon buttons
- `:where()` keeps specificity at 0 so themes can still override

## Test plan
- [ ] Insert a core/button block on a frontend page
- [ ] Hover — confirm subtle opacity dim + 1px upward lift
- [ ] Active state returns to baseline transform
- [ ] Theme-defined button hover styles still take precedence (specificity check)

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -699,6 +699,9 @@ class Plugin {
 		// Apply global default hover animation to Form Builder submit buttons.
 		add_filter( 'render_block_designsetgo/form-builder', array( $this, 'apply_default_form_button_hover' ), 10, 2 );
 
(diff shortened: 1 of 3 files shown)


jnealey-godaddy:
Closing — auto-applying the icon-button hover animation to core buttons doesn't produce a visible effect (theme's own hover bg sits in the same place, and even the icon-button's animation is essentially invisible without a user-set hover color). Will revisit as a proper core/button extension if/when needed.

PR #386, opened 2026-04-24
feat(dynamic-tags): Elementor-style dynamic tags system

jnealey-godaddy:
## Summary

Adds a full Elementor-style Dynamic Tags subsystem so authors can bind block content to post/site/archive/user fields or custom fields (ACF, Meta Box, Pods, JetEngine) without writing code. Built on WordPress's native Block Bindings API plus a thin metadata layer that gives us the picker UI, field discovery, and live preview that core doesn't ship yet.

### What ships

**1. A showcase block** — `designsetgo/dynamic-image`
  - Server-rendered via `render.php`
  - Source picker (featured image, site logo, ACF image, post-meta image, author avatar, …)
  - Size, focal point, aspect ratio, object-fit, fallback image, link-wrap

**2. Extension for core blocks** (`src/extensions/dynamic-tags/`)
  - Adds a per-attribute "⚡ Connect" button in the Inspector for every bindable attribute on `core/paragraph`, `core/heading`, `core/image`, `core/button`, `core/post-date`
  - Writes directly to native `attributes.metadata.bindings` so WP core renders through its own pipeline

**3. ~22 intrinsic binding sources**
  - Post: title, excerpt, date, modified-date, permalink, id, type, author-name, author-url, author-avatar-url, featured-image (url/id/alt/width/height/title/caption via `subkey`), taxonomy
  - Site: title, tagline, url, logo
  - Archive: title, description, url
  - User: current-user-name, current-user-avatar, current-user-url, current-user-email (opt-in via filter)

**4. Extension to existing custom-field sources**
  - ACF / Meta Box / Pods / JetEngine sources now accept `args.subkey` so image-type fields (which these plugins return as arrays) can resolve to `url` / `id` / `alt` / etc. through the scalar-only Bindings API

**5. REST API** (`/wp-json/designsetgo/v1/dynamic-tags/…`)
  - `/sources` — catalog with labels, groups, return types, arg schemas
  - `/fields` — discovered fields for a given source + post type
  - `/preview` — resolves the exact value `render.php` will emit, so the editor preview and the frontend can't drift

**6. Reusable React picker**
  - `src/components/DynamicTagPicker/` exposes `DynamicTagPicker`, `DynamicTagButton`, and three data hooks

### Architecture notes

- All new sources register via the pre-existing `designsetgo_register_bindings_source()` helper — they inherit post-password, viewability, and protected-meta gates for free.
- The Dynamic Image block does **not** render through `core/image` bindings. The scalar-only constraint of the Bindings API would force three round-trips (url/id/alt) and still not cover size variants or fallback. Instead, `render.php` calls `ImageResolver::resolve()` directly and the same resolver powers `/preview`, so editor and frontend always agree.
- The Dynamic Tags registry is kept *alongside* WP core's Bindings registry, not as a replacement — generic core bindings continue to work exactly as they do today.

### Out of scope (follow-ups)

- Binding DesignSetGo's own image-bearing blocks (card, media-text). The plan is to refactor those blocks onto the native Block Bindings API in a separate PR.
- Repeaters / arrays — WP core Bindings doesn't support them yet either.
- Global plugin-level fallback image (today: per-block only).

## Test plan

- [ ] `npm run build` — builds without errors (verified: passes with only pre-existing Sass `@import` deprecation warnings)
- [ ] PHP `-l` syntax check on every new/modified file — clean
- [ ] Insert Dynamic Image block → pick "Post featured image" → verify it renders the featured image on the frontend
- [ ] Insert Dynamic Image on a post with no featured image + a fallback → verify fallback renders
- [ ] Bind `core/heading` content to `designsetgo/post-title` via the Inspector → verify it serializes as `metadata.bindings` and renders the post title
- [ ] Bind `core/image` URL to `designsetgo/acf` with `subkey=url` on an ACF image field → verify the image renders
- [ ] Hit `/wp-json/designsetgo/v1/dynamic-tags/sources` logged out → expect 401; logged in as editor → expect JSON catalog
- [ ] Password-protect a post, set an image binding → verify `/preview` returns `{status: "unauthorized"}` and the frontend hides the image
- [ ] Verify `/preview` result for a source matches what `render.php` outputs on the frontend (parity check)
- [ ] Check `current-user-email` source is absent unless the `designsetgo_dynamic_tags_allow_email` filter is set to true

https://claude.ai/code/session_01HXWvVyvftute8yXbAtc53Y

---
_Generated by [Claude Code](https://claude.ai/code/session_01HXWvVyvftute8yXbAtc53Y)_

### DIFF
--- a/includes/abilities/info/class-list-dynamic-tag-sources.php
+++ b/includes/abilities/info/class-list-dynamic-tag-sources.php
@@ -1,0 +1,178 @@
+<?php
+/**
+ * List Dynamic Tag Sources Ability.
+ *
+ * Returns the catalog of registered DesignSetGo Dynamic Tag sources —
+ * post / site / archive / user fields plus custom-field providers
+ * (ACF, Meta Box, Pods, JetEngine). AI agents use this to discover
+ * which `metadata.bindings.<attr>.source` values are available before
+ * writing bindings to a block via update-block.
+ *
+ * @package DesignSetGo
+ * @subpackage Abilities
+ * @since 2.2.0
+ */
+
+namespace DesignSetGo\Abilities\Info;
+
+use DesignSetGo\Abilities\Abstract_Ability;
+use DesignSetGo\Blocks\DynamicTags\Registry;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * List Dynamic Tag Sources ability.
+ */
+class List_Dynamic_Tag_Sources extends Abstract_Ability {
+
+	/**
+	 * Get ability name.
+	 *
+	 * @return string
+	 */
+	public function get_name(): string {
+		return 'designsetgo/list-dynamic-tag-sources';
+	}
+
+	/**
+	 * Get ability configuration.
+	 *
+	 * @return array<string, mixed>
+	 */
+	public function get_config(): array {
+		return array(
+			'label'               => __( 'List Dynamic Tag Sources', 'designsetgo' ),
+			'description'         => __( 'Returns the catalog of Dynamic Tag binding sources (post/site/archive/user fields, ACF, Meta Box, Pods, JetEngine). Each entry includes the source slug, label, group, return types, and arg schema. Use the slug as `metadata.bindings.<attribute>.source` when writing bindings via update-block.', 'designsetgo' ),
+			'category'            => 'info',
+			'input_schema'        => $this->get_input_schema(),
+			'output_schema'       => $this->get_output_schema(),
+			'permission_callback' => array( $this, 'check_permission_callback' ),
+			'show_in_rest'        => true,
+			'keywords'            => array( 'binding', 'dynamic', 'token', 'tag', 'source', 'data' ),
+			'annotations'         => array(
+				'readonly'     => true,
+				'instructions' => 'Read-only catalog. Filter by `returns` (text|image|url|number|date) to narrow to sources compatible with the bound attribute, or by `group` to scope to one family. To bind a source, write `attributes.metadata.bindings.<attr> = { source: "<slug>", args: { ... } }` via update-block.',
+			),
+		);
+	}
+
+	/**
+	 * Capability gate — same baseline as the editor.
+	 *
+	 * @return bool
+	 */
+	public function check_permission_callback(): bool {
+		return current_user_can( 'edit_posts' );
+	}
+
+	/**
+	 * Input schema.
+	 *
+	 * @return array<string, mixed>
+	 */
+	private function get_input_schema(): array {
+		return array(
+			'type'                 => 'object',
+			'properties'           => array(
+				'returns' => array(
+					'type'        => 'string',
+					'enum'        => array( 'text', 'image', 'url', 'number', 'date' ),
+					'description' => __( 'Filter to sources that can return the given type.', 'designsetgo' ),
+				),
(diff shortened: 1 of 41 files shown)


jnealey-godaddy:
@claude review the last few commits

jnealey-godaddy:
Self-review of the recent commits (`a19be74` → `6d35df7`):

### Things that work well

- **Theme 3 inspector migration** of `dynamic-image` is clean — single `DsgoInspectorPanel`, one `DsgoInspectorPanel.Item` per attribute, `resetAll` covers every default. Inspector-IA test suite registers it and 350/350 tests pass.
- **Live editor preview** via `registerBlockBindingsSource` (`register-bindings-sources.js`) is the right pattern. Reading from `core/editor` for post-* and `core/core-data` for site-* makes the canvas reflect unsaved edits without a REST hop.
- **Abilities API entry** auto-registers via the existing `info/` glob — `designsetgo/list-dynamic-tag-sources` exposes the full registry to AI agents with a sensible filter shape (`returns`, `group`).

### Things worth a second look

1. **Extension disable is JS-only.** `Bootstrap` always instantiates server-side; `enabledExtensions` only gates the JS UI. So if a site disables `dynamic-tags`, the picker disappears but PHP keeps resolving any bindings already saved into posts. That's probably the *right* behaviour (don't break existing content) but it's worth a docblock note. Alternative: also gate `Bootstrap` instantiation behind the same `should_load_extension` filter, with a fallback that still resolves saved bindings.

2. **Live preview is silent in the Site Editor.** All my JS resolvers read from `core/editor` which doesn't exist in FSE / template editing contexts. Acceptable for v1 (the placeholder still shows), but worth a TODO. A future fix would `select(coreStore).getEntityRecord('postType', postType, postId)` for parity.

3. **`require()` inside an ES module** in `extensions/dynamic-tags/index.js` works under webpack but is unusual and would trip stricter ESLint configs. Cleaner pattern: top-level `import` of a registration function and call it conditionally.

4. **Error swallow on already-registered binding sources.** `try/catch (e) {}` in `register-bindings-sources.js` hides any future API mismatch. `getBlockBindingsSource(slug)` exists on `@wordpress/blocks` — preferring an existence check over swallowing all errors would be safer.

5. **`post-id` returns the string `"0"` for never-saved posts.** Probably should return empty string when `getCurrentPostId()` is 0, otherwise authors see a bare `0` in the canvas before save.

6. **Modal frame height math** — I just fixed a regression where `height: auto + max-height` pushed the footer below the fold. Now uses `min(640px, 80vh)`. Worth confirming on a small viewport (≤ 700px) that the source list still scrolls cleanly.

7. **Bindings JS resolver projects one scalar across every bound attribute.** `buildGetValues` returns the same value for every attribute key in `bindings`. For text-typed sources bound to text-typed attributes (the only path the picker allows) this is correct, but a defensive guard against multi-attribute `core/image` bindings would be cheap.

### Out of scope but tracked

- E2E failure on `query-block.spec.js` (`frontend renders posts list and emits ItemList schema`) is a `locator.click` timeout against the publish button blocked by a `.components-modal__screen-overlay`. The query block doesn't pull in my extension and my picker modal only opens on explicit user action — looks like a flaky pre-existing test, will reverify after the latest commit's CI run.
- PHPUnit tests for the new REST controller and Registry haven't landed yet (Copilot review flagged this earlier). Tracking as a follow-up PR.

---

PR #387, opened 2026-04-24
2.1.0 prep: Dynamic Query/Section authoring + security & perf hardening

jnealey-godaddy:
## Summary

Bundles release-prep work from a sibling session (CHANGELOG/readme rewrite for 2.1.0, Section block "clears default padding when nested" rule) with the security & performance hardening pass from the recent deep-dive audit.

## Hardening (audit follow-ups)

| ID | Fix | Files |
|---|---|---|
| H1 | `safeQueryAll()` wrapper around `document.querySelectorAll(customSelector)` so an admin typo can no longer throw `DOMException` and kill all frontend JS | `src/utils/sticky-header.js` |
| H2 | `esc_sql()` defence-in-depth on 5 raw `TRUNCATE` / `DROP TABLE` / `SELECT COUNT(*)` statements that concatenate the table name | `includes/blocks/class-query-filter-index{,-rebuilder,-cli}.php` |
| H3 | Validate `Content-Type` + `typeof data.html === 'string'` before swapping `region.innerHTML` with REST output | `src/blocks/query/view.js` |
| M1 | Production builds emit no source maps (`devtool: false` in prod); `*.map` added to `.distignore` | `webpack.config.js`, `.distignore` |
| M2 | New `Settings::sanitize_css_selector()` rejects `<>'"`, `javascript:`, `expression(`, `url(`, `@import`; routed from both settings schema and `wp_localize_script` | `includes/admin/class-settings.php`, `includes/class-sticky-header.php` |
| M3 | `phpcs:ignore` annotation on own-build `file_get_contents()` | `includes/class-assets.php` |
| M4 | Tighter CSS-value allowlist (numeric+unit, `var(...)`, hex, rgb/hsl, keyword, font-family list); invalid values silently dropped | `includes/admin/class-global-styles.php` |
| L2/L3 | `$_request` rename + dead `$saved_styles` parameter removed from 4 methods/call sites | `includes/admin/class-global-styles.php` |
| M5 | Stricter PHP fallback rate limit (3 → 2 / 60s) for orphaned submissions; clearer comment documenting the layered anonymous defences | `includes/blocks/class-form-handler.php` |
| L4 | `error_log()` wrapped in `WP_DEBUG` guard + `phpcs:ignore` to match codebase convention | `includes/admin/class-block-migrator.php` |
| L5 | Inline-style status concatenation replaced with `dsgo-submission-status` BEM class + once-per-page `<style>` block | `includes/blocks/class-form-submissions.php` |
| L6 | New `File_Manager::fs_put_contents()` / `fs_delete()` helpers using `WP_Filesystem` (with explicit fallback); 10 call-sites migrated. Fixes silent write failures on managed hosts (WP Engine, Kinsta, Pantheon) | `includes/llms-txt/class-{file-manager,controller,rest-controller}.php` |

## Other agent's release-prep work bundled in

- `CHANGELOG.md` — restructured the Unreleased entry into a fully fleshed-out 2.1.0 changelog (Dynamic Query family, Dynamic Tags, Block Bindings sources, conditional visibility, Markdown content negotiation, hover effects, grid column toolbar, form persistence, Themes 1–6 editor UX rollout, etc.).
- `readme.txt` — pulled WordPress.org changelog forward to match.
- `src/blocks/section/style.scss` — Section now clears its default padding automatically when nested inside another Section.

## Verification

- `npm run clean:build && npm run build` — succeeds, **0 `.map` files** emitted in `build/`.
- `phpcs --standard=phpcs.xml --warning-severity=0` on all 13 changed PHP files — **13/13 pass, 0 errors**.
- `wp-scripts lint-js` on the two changed JS files — clean.

Pre-existing lint warnings in unrelated files (visibility, dynamic-tags, style-binding extensions, slider editor.scss) were not touched.

## Test plan

- [ ] Sticky header: set `custom_selector` to `[invalid` in Settings → DesignSetGo, confirm frontend JS still runs (header silently falls back).
- [ ] Sticky header: set `custom_selector` to `<script>` in Settings, confirm it's rejected and persisted as empty.
- [ ] Dynamic Query: load a page with a Query block + filters, change filters, confirm the region updates without console errors.
- [ ] Dynamic Query: in WP-CLI run `wp dsgo query index status` / `rebuild` / `drop` to confirm the table-name DDL still works after `esc_sql()`.
- [ ] Form builder: submit a form anonymously 3× in 60s and confirm the new 2/60s fallback rate-limits correctly when no per-block rateLimitCount is configured.
- [ ] Form submissions admin: open a submission, confirm the email-sent / not-sent banner renders with the new BEM class (no inline-style attribute).
- [ ] llms.txt: enable the feature on a managed-host install (or with a non-writable ABSPATH), confirm `llms.txt` and `llms-full.txt` write successfully via WP_Filesystem.
- [ ] Section: nest a Section inside another Section, confirm the inner one drops its default padding.
- [ ] Build artifact: confirm `build/` ships zero `.map` files after a clean production build.

### DIFF
--- a/.distignore
+++ b/.distignore
@@ -147,3 +147,9 @@ npm-debug.log*
 # ===================
 languages/*.po
 languages/*.pot
+
+# ===================
+# Source Maps (never ship to production)
+# ===================
+*.js.map
+*.css.map
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Patterns: removed leftover `id="contact-professional"` on the form-builder root in starter patterns; aligned form block markup with current `save.js` output.
 - Abilities API add-block output round-tripped through `save()` to prevent block validation failures (#355).
 - Abilities JSON Schema — inline `required:true` migrated to JSON Schema compliant form (#352).
+- llms.txt generation now writes reliably on managed hosts (WP Engine, Kinsta, Pantheon) that don't define FTP constants — file writes go through the WordPress filesystem API with a safe fallback for environments where it's unavailable.
+- Sticky header: a typo in the Settings → DesignSetGo custom selector field no longer throws an exception that breaks all frontend JavaScript; an invalid selector silently falls back to the default header detection.
 
 ### Removed
 - **Visual Revision Comparison** — Removed in favor of WordPress 7.0's native visual diff for revisions. Also removed the associated admin page, block differ, revision renderer, REST endpoints (`/designsetgo/v1/revisions/*`), and settings (`revisions.enable_visual_comparison`, `revisions.default_to_visual`).
@@ -82,6 +84,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Draft Mode REST permission callbacks** — Added nonce verification to all Draft Mode REST routes.
 - **Form submissions** — Redirect URLs normalized and validated against an allowlist of safe protocols before navigation.
 - **Dynamic Query style bindings** — CSS value injection blocked; dangerous CSS functions and protocols cannot be passed through a binding.
+- **Global Styles sanitization** — Stored Global Styles values are now validated against a CSS-value allowlist (numeric+units, `var()`/`calc()`/`clamp()`/`min()`/`max()`, hex/rgb/hsl colors, named keywords, font-family lists). Every functional form rejects `url(`, `expression(`, `javascript:`, `<`, and `;` regardless of the wrapping function.
+- **Sticky header custom selector** — Settings input now rejects HTML angle brackets, `javascript:`, `expression(`, `url(`, and `@import` patterns before the value reaches the frontend.
 
 ### Changed
 - `designsetgo/post-meta` and `designsetgo/acf` binding sources accept an optional `scope` arg — defaults to `'self'`; existing bindings unchanged.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,126 @@ All notable changes to the DesignSetGo plugin will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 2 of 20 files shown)


jnealey-godaddy:
## Review feedback addressed in a0e54221

Both bots flagged useful issues — thanks. All addressed:

| Reviewer | Finding | Resolution |
|---|---|---|
| Copilot | `view.js:791` Content-Type check is case-sensitive | `.toLowerCase()` before `.includes('application/json')` |
| Copilot | `class-global-styles.php:636` rejects `calc()`/`clamp()`/`min()`/`max()` (theme.json uses these heavily) | Folded these into a unified functional-CSS branch in `is_valid_css_value()` |
| Copilot + Claude | `class-settings.php` rejects `"` and `'`, breaking `[data-foo="bar"]` attribute selectors | Quotes now allowed; `<>`, `javascript:`, `expression(`, `url(`, `@import` still rejected |
| Claude | `class-global-styles.php:617` `var()` regex rejects nested `var(--a, var(--b))` fallbacks | Same unified functional-CSS branch handles `var()` with arbitrary nested args, blocked only by the shared sink list (`url(`, `expression(`, `javascript:`, `<`, `;`) |
| Claude | `class-global-styles.php:627` `rgb()`/`hsl()` allow `url(`/`expression(` in args | Now blocked via the shared sink list |
| Claude | `class-query-filter-index{,-rebuilder,-cli}.php` — `esc_sql()` is for string delimiters, not bare identifiers | Switched all 5 sites to `$wpdb->prepare('… %i', $table)` (WP 6.2+); plugin requires WP 6.7+ so this is safe |
| Claude | `class-file-manager.php:66` fallback comment undersells managed-host scenario | Updated comment — managed hosts (WP Engine, Kinsta, Pantheon) without FTP constants are now called out as the primary path |

### Skipped (with rationale)

- **`readme.txt` stable tag `2.0.51`** (Copilot): this is the WP.org convention — stable tag stays at the currently-shipped version while a new one is being prepared in trunk. The Claude review explicitly confirmed this is correct for an unreleased branch targeting 2.1.0. The other agent set this deliberately.

### Verification

- `phpcs --warning-severity=0` — clean on the 6 changed files
- `wp-scripts lint-js` — clean on `view.js`
- `npm run build` — clean, 0 `.map` files

PR #391, opened 2026-04-27
fix: defer DynamicTags default group registration to after_setup_theme

ncimbaljevic-godaddy:
## Description
Calling __() in register_default_groups() at plugin load time triggers the _load_textdomain_just_in_time notice on WordPress 6.7+ because translations are not available before after_setup_theme. Defer group registration to after_setup_theme, which still runs well before sources register at init priority 6.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement

## Related Issue

Closes #

## Changes Made

- Defer DynamicTags\Registry::register_default_groups() from plugin load time to after_setup_theme hook, fixing _load_textdomain_just_in_time notice on WP 6.7+
- Change register_default_groups() visibility from private to public so it can be called by add_action()

## Testing

- [ ] Tested on frontend
- [ ] Tested responsive behavior (mobile/tablet/desktop)
- [ ] No console errors
- [ ] No PHP errors

## Checklist

- [ ] My code follows the [WordPress Coding Standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/)
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have updated documentation as needed
- [ ] My changes generate no new warnings or errors
- [ ] I have tested on WordPress 6.4+
- [ ] I have followed the patterns in [CLAUDE.md](.claude/CLAUDE.md)
- [ ] All files are under 300 lines (if applicable)
- [ ] I have added JSDoc comments to new functions
- [ ] Accessibility: WCAG 2.1 AA compliant
- [ ] Security: All user input is validated and sanitized
- [ ] Internationalization: All user-facing strings use `__()` or `_e()`

### DIFF
--- a/includes/abilities/class-abilities-registry.php
+++ b/includes/abilities/class-abilities-registry.php
@@ -256,6 +256,14 @@ class Abilities_Registry {
 			return;
 		}
 
+		// On WP 6.9 the wp_abilities_api_categories_init hook fires before
+		// init, so __() calls here trigger a _load_textdomain_just_in_time
+		// notice.  Defer to init when called too early.
+		if ( ! did_action( 'init' ) ) {
+			add_action( 'init', array( $this, 'register_ability_categories' ) );
+			return;
+		}
+
 		wp_register_ability_category(
 			'info',
 			array(
@@ -289,6 +297,14 @@ class Abilities_Registry {
 	public function register_abilities(): void {
 		// Check if the Abilities API is available.
 		if ( ! class_exists( 'WP_Ability' ) ) {
+			return;
+		}
+
+		// Defer to init when the abilities hook fires too early (WP 6.9)
+		// to avoid _load_textdomain_just_in_time notice from __() inside
+		// each ability's get_config().
+		if ( ! did_action( 'init' ) ) {
+			add_action( 'init', array( $this, 'register_abilities' ) );
 			return;
 		}
 
--- a/includes/abilities/class-abilities-registry.php
+++ b/includes/abilities/class-abilities-registry.php
@@ -260,7 +260,7 @@ class Abilities_Registry {
 		// init, so __() calls here trigger a _load_textdomain_just_in_time
 		// notice.  Defer to init when called too early.
 		if ( ! did_action( 'init' ) ) {
-			add_action( 'init', array( $this, 'register_ability_categories' ) );
+			add_action( 'init', array( $this, 'register_ability_categories' ), 8 );
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


jnealey-godaddy:
# PR Review Summary

**fix: defer DynamicTags default group registration to after_setup_theme** — 3 files, +22/-4

The diagnosis is correct: `__()` called before `init` on WP 6.7+/6.9 triggers `_load_textdomain_just_in_time` notices. The fix works in the normal plugin-load flow, but two of the three changes have rough edges worth addressing before merge.

## Important

**1. Dynamic-tags deferral is missing a `did_action` guard** — `includes/blocks/dynamic-tags/class-dynamic-tags-registry.php:54`

```php
self::$instance = new self();
add_action( 'after_setup_theme', array( self::$instance, 'register_default_groups' ) );
```

If `Registry::instance()` is ever first called *after* `after_setup_theme` fires, the action queues against a hook that will never re-fire, leaving `$this->groups` empty. In the current codebase `Bootstrap` is constructed from `Plugin::init()` at plugin load (well before `after_setup_theme`), so this is latent — but the abilities fix in this same PR uses the more defensive pattern. Mirror it for consistency:

```php
if ( did_action( 'after_setup_theme' ) ) {
    self::$instance->register_default_groups();
} else {
    add_action( 'after_setup_theme', array( self::$instance, 'register_default_groups' ) );
}
```

**2. Pattern inconsistency between the two registries** — abilities uses a `did_action('init')` guard; dynamic-tags uses a bare `add_action`. Same problem, two different solutions in one PR. Pick one.

**3. Ordering risk in abilities deferral** — `includes/abilities/class-abilities-registry.php` (both new guard blocks)

Both `register_ability_categories` and `register_abilities` reschedule to `add_action('init', ...)` at default priority 10. The original hooks (`wp_abilities_api_categories_init` / `wp_abilities_api_init`) guarantee categories fire first. When deferred, that ordering relies only on FIFO at priority 10. Make the contract explicit:

```php
add_action( 'init', array( $this, 'register_ability_categories' ), 9 );
```

## Suggestions

**4. Visibility change `private` → `public` on `register_default_groups()`** — exposes an internal seeding step as a public API surface just to satisfy `add_action`'s callable. Either keep it `private` and dispatch via a closure, or accept the new surface and note it's intentional in the docblock.

**5. Unrelated comment removal in `includes/class-plugin.php:707-712`** — the deleted `load_plugin_textdomain` note isn't part of the JIT-notice fix. The comment was factually correct and useful context. Either restore it or split the cleanup into its own commit/PR.

**6. No logging when deferral can't help** — if a future codepath hits the singleton too late, the registries silently end up empty. A `_doing_it_wrong()` call in that branch would give developers a traceable signal instead of an empty dropdown.

## Strengths

- Correct root-cause diagnosis; the `__()`-before-`init` issue is real on 6.7+ and worsens on 6.9.
- The comment in the dynamic-tags registry documents the *why* and the timing relationship to `init` priority 6 sources — exactly what's needed.
- Small, focused diff.

## Recommended action

Address (1) and (2) — make dynamic-tags use the same `did_action` guard as abilities. Decide on (3), (4), (5). (6) is optional polish.

PR #404, opened 2026-05-13
fix(abilities): allow empty input on get-global-css GET

jnealey-godaddy:
## Summary
- `designsetgo/get-global-css` returned **400 `ability_invalid_input`** on every GET because the schema declared `type: object` with no default and core's Abilities run controller passes `null` when no `?input=` query param is present.
- Adding `'default' => array()` lets `WP_Ability::normalize_input()` supply `[]` before validation runs — satisfies the schema and the `array` type-hint on `execute()`.
- Verified via the PR-395 smoke test that found the bug: GET with no input, write-then-read round-trip, and the cleared state all return 200 with the expected payload.

## Test plan
- [x] `phpunit --filter Abilities_Settings_Test` — 3/3 (regression: null → `[]` normalization, empty-state read, write→read round-trip)
- [x] `phpunit --filter Abilities` — 100/100, no regressions in adjacent ability tests
- [x] `phpcs --standard=phpcs.xml` clean on changed files
- [x] Manual REST exercise from `wp-admin` against running wp-env: `GET /wp-json/wp-abilities/v1/abilities/designsetgo/get-global-css/run` returns 200 with `{success, css, stylesheet, post_id}`

## Notes
- The `<script>...</script>` value passes through `update-global-css` unchanged. That mirrors WP core's Customizer Additional CSS behaviour — the value is only ever rendered inside `<style>`, so it's inert — and is therefore not changed here.
- A `_doing_it_wrong` notice on `WP_Ability::__construct` for an unrelated `keywords` property fires once per process when the registry is first touched. The new test suite declares it expected on the first `setUp` only; the underlying `keywords`/`meta` migration is its own follow-up.

### DIFF
--- a/includes/abilities/class-abstract-ability.php
+++ b/includes/abilities/class-abstract-ability.php
@@ -96,6 +96,15 @@ abstract class Abstract_Ability {
 			unset( $config['annotations'] );
 		}
 
+		// Move keywords into meta as well. WP_Ability::__construct emits a
+		// _doing_it_wrong notice for any unknown top-level property, and
+		// `keywords` is one — historically we declared it at the top level
+		// to match the registration shape from earlier prototypes.
+		if ( isset( $config['keywords'] ) ) {
+			$config['meta']['keywords'] = $config['keywords'];
+			unset( $config['keywords'] );
+		}
+
 		wp_register_ability( $this->get_name(), $config );
 	}
 
--- a/includes/abilities/settings/class-get-global-css.php
+++ b/includes/abilities/settings/class-get-global-css.php
@@ -47,6 +47,11 @@ class Get_Global_CSS extends Abstract_Ability {
 				'type'                 => 'object',
 				'properties'           => new \stdClass(),
 				'additionalProperties' => false,
+				// REST GETs without `?input=...` arrive as null. The Abilities
+				// API replaces null with the schema default before validation,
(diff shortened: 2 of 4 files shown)


jnealey-godaddy:
Addressed in 82d2610c:

- **Medium / fragile cleanup** — moved CSS reset into `tear_down()`, so a failing mid-test assertion can no longer leak global Additional CSS state.
- **Low / `post_id` empty-state assertion** — added `assertArrayHasKey('post_id')` + `assertNull($result['post_id'])`; the round-trip test now also asserts `isInt` after a write.
- **Low / permission denial** — added `test_permission_denied_for_subscriber` covering both `get-global-css` and `update-global-css`, mirroring the pattern in `Abilities_Security_Test`.
- **`_doing_it_wrong` root cause** — fixed at the source: `Abstract_Ability::register()` now migrates `keywords` into `meta` (same as it already does for `show_in_rest` and `annotations`). The `setExpectedIncorrectUsage` workaround is gone.

CI failures also addressed in the same commit:
- PHP 8.0–8.3 Lint: Yoda-condition fixes on `includes/llms-txt/class-generator.php:142,245` (pre-existing, from PR #403).
- WordPress 6.7 / PHP 8.0: skip with `markTestSkipped` when `wp_get_ability` is unavailable (Abilities API ships in 6.9).

PR #424, opened 2026-06-23
fix: harden Medium security findings S1–S6

ncimbaljevic-godaddy:
## Summary

Addresses all six **Medium** findings from a security audit of the plugin (no Critical/High exploitable vulnerabilities were found). All fixes are server-side PHP hardening — no block attribute schema or HTML changes, so **no block deprecations are required**.

| # | Finding | Fix |
|---|---------|-----|
| **S1** | Turnstile secret key leaked via REST GET `/settings` | Redact `integrations.turnstile_secret_key` to a placeholder on read; strip the placeholder on write so the stored secret is preserved (`write_only_secret_keys()` map) |
| **S2** | Mutating REST routes lacked nonce verification | Verify `X-WP-Nonce` on the 4 llms-txt POST routes (`check_admin_permission()`) and query template export/import (`verify_nonce()`) |
| **S3** | `user_url` returned without scheme validation | `current-user-url` dynamic-tags source now passes through `esc_url_raw(…, ['http','https'])` |
| **S4** | `custom_css_output` filter result echoed without re-sanitization | Extracted `strip_dangerous_css()` and re-run it on the filter result before echo |
| **S5** | select/checkbox/hidden fields lacked server-side allowed-value enforcement | New `get_form_field_value_constraints()` + `$allowed` param on `validate_field()` enforce membership/constants from the server-parsed block |
| **S6** | Incomplete object-injection guard in Draft Mode | Recursive `contains_object()` replaces flat `is_object()` in both meta-copy paths |

All affected REST routes are called from the editor/admin via `@wordpress/api-fetch`, which auto-injects `X-WP-Nonce`, so the nonce checks don't break the UI.

The two **Low** finding clusters (S7–S11) are intentionally out of scope for this PR.

## Test plan

- `composer lint` (phpcs over `includes/`) — clean
- `vendor/bin/phpstan analyse --memory-limit=1G` — no errors
- Full PHPUnit suite via wp-env — **804 passing**, 0 failures (6 pre-existing env skips)
- Added regression tests for S1 (redaction round-trip), S4 (output-filter re-sanitization), S5 (allowed-value enforcement), S6 (nested-object guard); updated query-template / llms-txt tests to send a valid nonce

### DIFF
--- a/includes/admin/class-draft-mode.php
+++ b/includes/admin/class-draft-mode.php
@@ -365,8 +365,9 @@ class Draft_Mode {
 
 			foreach ( $values as $value ) {
 				$unserialized = maybe_unserialize( $value );
-				// Reject PHP objects to prevent object injection.
-				if ( is_object( $unserialized ) ) {
+				// Reject PHP objects (including objects nested inside arrays)
+				// to prevent object injection.
+				if ( self::contains_object( $unserialized ) ) {
 					continue;
 				}
 				add_post_meta( $target_id, $key, $unserialized );
@@ -421,8 +422,9 @@ class Draft_Mode {
 
 			foreach ( $values as $value ) {
 				$unserialized = maybe_unserialize( $value );
-				// Reject PHP objects to prevent object injection.
-				if ( is_object( $unserialized ) ) {
+				// Reject PHP objects (including objects nested inside arrays)
+				// to prevent object injection.
+				if ( self::contains_object( $unserialized ) ) {
 					continue;
 				}
 				add_post_meta( $original_id, $key, $unserialized );
@@ -461,5 +463,31 @@ class Draft_Mode {
 				}
 			}
 		}
+	}
+
+	/**
+	 * Recursively determine whether a value is, or contains, a PHP object.
+	 *
+	 * A plain is_object() check misses objects nested inside arrays (e.g. a
+	 * serialized payload like `a:1:{i:0;O:8:"stdClass":0:{}}`), which would
+	 * still trigger object injection when stored and later unserialized.
+	 *
+	 * @param mixed $value Value to inspect (already passed through maybe_unserialize()).
+	 * @return bool True if the value is an object or any nested array element is.
+	 */
+	private static function contains_object( $value ) {
+		if ( is_object( $value ) ) {
+			return true;
+		}
+
+		if ( is_array( $value ) ) {
+			foreach ( $value as $item ) {
+				if ( self::contains_object( $item ) ) {
+					return true;
+				}
+			}
+		}
+
+		return false;
 	}
 }
--- a/includes/admin/class-settings.php
+++ b/includes/admin/class-settings.php
@@ -23,6 +23,16 @@ class Settings {
 	const OPTION_NAME = 'designsetgo_settings';
 
 	/**
+	 * Placeholder returned in place of a stored write-only secret on read.
+	 *
+	 * The REST GET endpoint redacts write-only secrets to this value so they
+	 * never reach the browser. The write path treats an incoming value equal
+	 * to this placeholder as "unchanged" and preserves the stored secret. A
+	 * real secret will never equal this string.
+	 */
+	const REDACTED_PLACEHOLDER = '••••••••';
+
+	/**
 	 * Constructor
 	 */
 	public function __construct() {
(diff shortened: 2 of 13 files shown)


jnealey-godaddy:
## PR Review: fix: harden Medium security findings S1–S6

**Verdict: Approve with minor comments**

A well-scoped, server-side-only PHP hardening PR (13 files, +666/−25) fixing six Medium-severity audit findings. No block/HTML/attribute changes → no deprecations needed. CI is green and every finding ships with regression tests.

### What's good
- **S1 single-source secret map** — `write_only_secret_keys()` drives both the read-path redaction in `redact_secrets()` and the write-path strip in `sanitize_settings()`; adding a future secret touches exactly one place.
- **S4 defense-in-depth** — `strip_dangerous_css()` runs after *both* the `custom_css_sanitize` filter and the `custom_css_output` filter, guarding against compromised/careless third-party filter callbacks at both injection surfaces.
- **Strong test coverage** — dedicated regression tests per finding; integration tests updated with valid nonces, proving the new checks don't silently break the existing suite.

### Issues to address

**Important**
1. **Missing nonce-*rejection* tests (S2)** — every test that exercises the nonce-protected routes supplies a *valid* nonce. Nothing asserts that a request with a missing/invalid nonce returns `401` from the new `verify_nonce()` / `check_admin_permission()`. A regression that removes the guard would be invisible to CI. The whole point of S2 is to *reject* CSRF — that path should be tested explicitly. Add at least two rejection tests (one for query template export/import, one for a llms-txt POST route).

   Example:
   ```php
   public function test_export_rejects_missing_nonce() {
       wp_set_current_user( self::factory()->user->create( array( 'role' => 'editor' ) ) );
       $post_id = self::factory()->post->create();
       $request = new WP_REST_Request( 'GET', '/designsetgo/v1/query/template' );
       // Intentionally omit X-WP-Nonce header
       $request->set_query_params( array( 'post_id' => $post_id, 'query_id' => 'abc' ) );
       $response = rest_do_request( $request );
       $this->assertSame( 401, $response->get_status() );
   }
   ```

2. **Nonce on a GET (READABLE) export route is non-standard** — `permission_export()` requires `X-WP-Nonce` on a GET request. GET is idempotent and not a CSRF vector by definition; the real guard is the existing `current_user_can( 'edit_post', $post_id )` check. This also diverges from `check_read_permission()` in `class-settings.php`, which does not require a nonce on GET. Any legitimate consumer that doesn't auto-inject the nonce (CLI tool, separate admin page) would get an unexpected 401. Either remove the nonce check from `permission_export()` (leaving the capability check) or add a comment documenting the deliberate "editor-only" policy.

**Optional**
- `google_maps_api_key` is not in `write_only_secret_keys()` — if its browser exposure is intentional (Maps JS API), add a comment; otherwise consider making it write-only too.
- `contains_object()` (`class-draft-mode.php`) has unbounded recursion — a simple depth cap (~20 levels) would make the defensive intent explicit and eliminate the theoretical stack-overflow risk.
- 401 vs 403 for nonce failure — cosmetic; matches the existing codebase pattern (`class-query.php`), so no change needed here, but worth noting for a future hardening pass.
- `REDACTED_PLACEHOLDER` uses non-ASCII bullet chars — minor opcode-cache/encoding risk on some hosts; a pure-ASCII sentinel (e.g. `__REDACTED__`) would be equally collision-resistant.

### Bottom line
The fixes themselves are correct and surgically focused. The only material gap is the missing nonce-rejection tests (Issue #1) — worth adding before merge to `release/2.2.0` so the S2 guard is verifiable by CI.

ncimbaljevic-godaddy:
1. Nonce-rejection tests (S2) — added explicit 401 regression tests so the CSRF guards can't silently regress. Each uses a user that passes the capability check, so removing the nonce guard would fail the test:            
- QueryTemplateControllerTest::test_import_rejects_missing_nonce / test_import_rejects_invalid_nonce — query-template import POST → 401
- Test_LLMS_Txt::test_flush_cache_endpoint_rejects_missing_nonce / test_flush_cache_endpoint_rejects_invalid_nonce; llms-txt flush-cache POST → 401                                                                           

2. Nonce on the GET export route removed now. permission_export() now relies solely on the edit_post capability check, matching Settings::check_read_permission(). verify_nonce() still guards the mutating import POST. Updated the GET export tests to no longer send a nonce, so they double as proof the route works nonce-free.

3.  google_maps_api_key note: it's a public-facing key, so leaving it out of write_only_secret_keys() is correct (making it write-only would break the client-side Maps usage)

4. contains_object() recursion; capped at a new Draft_Mode::MAX_OBJECT_SCAN_DEPTH = 20. A payload nested deeper is treated as object-bearing (fail-closed), so the draft meta-copy skips it and there's no stack-overflow risk from a maliciously deep serialized value. Added a depth-cap regression test.

5. REDACTED_PLACEHOLDER; swapped the non-ASCII bullet string for a pure-ASCII sentinel (__DSGO_REDACTED__), byte-identical across encodings/opcode caches. It's referenced only via the constant, so the redact/strip round-trip is unchanged.

PR #445, opened 2026-07-07
feat(section): emit overlay class for style-kit overlay variations

jnealey-godaddy:
## Summary

Enables style kits to drive a section's background overlay through a block-style variation (`is-style-overlay-dark`, and any future `is-style-overlay-*`) instead of the `overlayColor` attribute.

Previously `designsetgo/section` emitted `dsgo-stack--has-overlay` — the class that renders the `::before` overlay — **only** when the `overlayColor` attribute was set. Moving overlay color into a style-kit class therefore produced no overlay. This change makes the block also emit `dsgo-stack--has-overlay` when an `is-style-overlay-*` variation is present on `className`, so the pseudo-element renders and the color is supplied by the variation's stylesheet.

Unblocks the section-overlay patterns (e.g. `benefits-cover-centered-grid`, `benefits-staggered-icon-cards`) that are migrating overlay color from `overlayColor`/inline styles to the `is-style-overlay-dark` class.

## Changes

- **New `hasOverlayStyleClass()` helper** (`src/blocks/section/utils/has-overlay-style.js`) — detects the whole `is-style-overlay-` family, so future overlay variations work with no further plugin change. Shared by edit + save to keep editor preview and saved markup byte-identical.
- **`save.js` / `edit.js`** — emit `dsgo-stack--has-overlay` when `overlayColor` **or** an overlay variation is present. In the variation case, no inline `--dsgo-overlay-color` is emitted (color comes from the variation's CSS).
- **`deprecated.js`** — new **v7** deprecation (all three parts: `isEligible` / `save` / `migrate`). Sections carrying an overlay variation but saved before this change **migrate silently** rather than showing "Attempt Recovery". `save()` reproduces the pre-change output so it also byte-matches on WP versions that validate the deprecation's save.

## Tests

- Save-level tests for the new overlay-class behavior (variation on/off, `overlayColor` still works, unrelated `is-style-*` ignored, no inline color var for the variation path).
- v7 deprecation tests, including an end-to-end silent-migration round-trip (`console.toHaveInformed()` + `block.isValid`).
- Updated existing deprecation-array destructuring indices for the new v7 entry.

Full suite: `8326 passed`. `eslint` clean, `npm run build` clean, overlay CSS confirmed present in the compiled section stylesheets.

## Note on scope

The pattern markup edits (removing `customOverlayColor` / `style.color.text` / inline colors and adding `is-style-overlay-dark`) live in the external style-kit repo, not this plugin. This PR is the plugin-side prerequisite that makes `is-style-overlay-dark` actually produce an overlay.

### DIFF
--- a/src/blocks/icon-button/editor.scss
+++ b/src/blocks/icon-button/editor.scss
@@ -59,9 +59,12 @@
 // CRITICAL: Check if PARENT container has the variable, not the button itself
 // CRITICAL: Use __inner wrapper to avoid triggering on outer padding areas
 // CRITICAL: Use direct child selector (>) to only affect direct children
+// See style.scss: the --has-hover-button selector adds Section style-variation
+// support alongside the inline-var (attribute) gate.
 .dsgo-flex[style*="--dsgo-parent-hover-button-bg"]:hover .dsgo-flex__inner > .dsgo-icon-button .dsgo-icon-button__wrapper,
 .dsgo-grid[style*="--dsgo-parent-hover-button-bg"]:hover .dsgo-grid__inner > .dsgo-icon-button .dsgo-icon-button__wrapper,
-.dsgo-stack[style*="--dsgo-parent-hover-button-bg"]:hover .dsgo-stack__inner > .dsgo-icon-button .dsgo-icon-button__wrapper {
+.dsgo-stack[style*="--dsgo-parent-hover-button-bg"]:hover .dsgo-stack__inner > .dsgo-icon-button .dsgo-icon-button__wrapper,
+.dsgo-stack--has-hover-button:hover .dsgo-stack__inner > .dsgo-icon-button .dsgo-icon-button__wrapper {
 	background-color: var(--dsgo-parent-hover-button-bg) !important;
 	transition: background-color 0.3s ease;
 }
--- a/src/blocks/icon-button/style.scss
+++ b/src/blocks/icon-button/style.scss
@@ -146,9 +146,14 @@
 // CRITICAL: Check if PARENT container has the variable, not the button itself
 // CRITICAL: Use __inner wrapper to avoid triggering on outer padding areas
 // CRITICAL: Use direct child selector (>) to only affect direct children
+// The `--has-hover-button` selector adds Section style-variation support
+// (is-style-hover-button-*) alongside the inline-var (attribute) gate. The class
+// is only emitted for that variation, so the !important override never clobbers
+// a button whose parent has no hover-button color set.
 .dsgo-flex[style*="--dsgo-parent-hover-button-bg"]:hover .dsgo-flex__inner > .dsgo-icon-button,
 .dsgo-grid[style*="--dsgo-parent-hover-button-bg"]:hover .dsgo-grid__inner > .dsgo-icon-button,
-.dsgo-stack[style*="--dsgo-parent-hover-button-bg"]:hover .dsgo-stack__inner > .dsgo-icon-button {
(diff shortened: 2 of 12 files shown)


jnealey-godaddy:
## Verified in a real WP 6.7 editor (wp-env), not just Jest

Ran the change end-to-end through the actual registered block (all plugin filters/supports applied), the editor UI, and the frontend:

- **New authoring** — a section with `is-style-overlay-dark` emits `dsgo-stack--has-overlay`, keeps the variation class, emits **no** inline `--dsgo-overlay-color`, and round-trips as valid. Confirmed on both the editor canvas and the published frontend.
- **`overlayColor` path** — still emits the overlay class + inline color var (unchanged).
- **Control** — a section with no overlay variation emits no overlay class.
- **Silent migration** — a page saved with the pre-change markup (variation class, no overlay class) loads in the editor with **no "Attempt Recovery" banner** (`isValid: true`), and after edit+save the frontend serves `dsgo-stack--has-overlay`.

### Bug caught by browser testing (fixed in d7c9785)

Jest alone gave a false pass. The real editor validates a deprecation's `save()` against stored HTML before migrating, and `v7` omitted block.json's `style` spacing-padding default — so `v7.save()` rendered without the padding, failed the byte-match, and old sections would **not** have migrated (they'd show "Attempt Recovery"). The original Jest fixture was built from `v7.save()` itself (tautological), so it never exercised this.

Fix: added the `style` default to `v7`'s attribute schema (mirrors block.json), and rebuilt the migration-test fixture from the real `save()` output (then stripped the overlay class). The rebuilt test fails without the fix — verified by temporarily reverting it — so it's now a genuine regression guard.

jnealey-godaddy:
## Hover colors — full parity with overlay (verified in real WP editor + frontend)

Extended variation support to all hover effects. Two mechanisms, by CSS nature:

**Section hover background** — additive `::after` overlay, so it's safe when unset. Made it render unconditionally with a transparent fallback → works from **any** variation that sets `--dsgo-hover-bg-color`, composes freely, **CSS-only, no deprecation**.

**Hover text / icon bg / button bg** — these override a base value with `!important`, so they must be gated to "only when set" (else transparent text/icons on every hover). Added **per-effect class-based activation**, mirroring the overlay pattern:

| Variation family | Block emits | Effect |
|---|---|---|
| `is-style-hover-text-*` | `dsgo-stack--has-hover-text` | section hover text color |
| `is-style-hover-icon-*` | `dsgo-stack--has-hover-icon` | child icon hover bg |
| `is-style-hover-button-*` | `dsgo-stack--has-hover-button` | child button hover bg |

Per-effect families (not one generic hover family) are deliberate — they keep the `!important` overrides from clobbering a base value when a variation sets only some hover vars. The CSS adds the class-gated selector *alongside* the existing inline-`[style*=…]` gate, so inspector-set hover still works unchanged. No new deprecation: the v7/chain byte-matches pre-hover markup and passthrough-migrates.

### Verified in the browser (real WP 6.7)
- Each family emits **exactly** its own activation class (no cross-activation).
- Variation-supplied hover text turns the section text on hover (`rgb(0,160,0)` from a stylesheet var) — impossible before.
- **Clobber-safety:** a plain section with its own text color is **unchanged** on hover.
- Variation-supplied hover-icon color applies to a real child icon on hover (`rgb(0,0,255)`).
- Old hover-variation content migrates silently (`isValid`, no "Attempt Recovery").

### Note for the style kit
Because these are per-effect families, a single (mutually-exclusive) block-style variation activates one hover effect. Overlay + hover **background** compose with any variation; hover **text/icon/button** each need their own `is-style-hover-{text,icon,button}-*` variation (or the kit can write the hover rule directly in its own CSS).

PR #448, opened 2026-07-07
feat(modal,scroll-slides): back hardcoded labels & overlay opacity with real attributes

jnealey-godaddy:
## Description

Continues the effort to eliminate hardcoded, uncontrollable values from block `save()` output. Reviewed the five reported items; three were genuine and are fixed here, two were already covered by WordPress core Border block-support (details under Additional Notes).

## Type of Change

## Related Issue

Closes #

## Changes Made

- **`designsetgo/scroll-slides` — overlay opacity control.** The overlay scrim opacity was hardcoded at `0.8` in `save.js`, `edit.js`, and `render.php` (query mode). Added an `overlayOpacity` attribute (default `80`) and an "Overlay Opacity" slider that appears once an overlay color is set. The `80` default reproduces the previous `0.8` output.
- **`designsetgo/modal` — dead `modalLabel` attribute fixed.** `save.js` read `attributes.modalLabel` but the attribute was never registered in `block.json`, so a custom dialog `aria-label` could never take effect and always fell back to `"Modal"`. Registered `modalLabel` and wired it to a new "Accessible Label" control.
- **`designsetgo/modal` — close-button label control.** The close button's `aria-label` (`"Close modal"`) was unconditionally hardcoded. Added a `closeButtonLabel` attribute plus a "Close Button Label" control; falls back to the previous string when empty. Editor preview mirrors the same fallback.
- Updated the unreleased `2.4.0` CHANGELOG section.

## Testing

- [x] Tested in WordPress editor
- [x] No console errors
- [x] No PHP errors

Verification performed:
- `npm run build` — compiles cleanly (only pre-existing asset-size warnings).
- `wp-scripts lint-js` on all changed files — clean.
- `php -l src/blocks/scroll-slides/render.php` — no syntax errors.
- `test:unit` modal suite — 33/33 pass.

## Backward compatibility

Every new attribute's default/empty value reproduces the prior markup byte-for-byte (`modalLabel: '' → "Modal"`, `closeButtonLabel: '' → "Close modal"`, `overlayOpacity: 80 → 0.8`), so **no block deprecations are required** and existing content validates unchanged.

## Additional Notes

Two of the five reported items were reviewed and left as-is because WordPress core Border block-support already provides the control:

- **`designsetgo/pill` border-radius** — the block registers `__experimentalBorder.radius`, and the `border-radius: 999px` fallback in CSS uses no `!important`. An author can override it (including setting `0` as the off switch) via the Border → Radius control; `render.php` already moves the border style group onto `.dsgo-pill__content` where the fallback lives. The original report itself noted this claim was overstated.
- **`designsetgo/slider` style-variation radius** — the block also registers `__experimentalBorder.radius`. The per-variation radii (`.dsgo-slider--classic { border-radius: 8px }`, etc.) are plain class selectors that any inline Border → Radius value overrides, including `0`.

Adding dedicated "disable radius" attributes to these two would duplicate core support and bake redundant values into saved markup, which runs counter to the goal of this cleanup series.

https://claude.ai/code/session_01ACY5FuAq6EyXS6tqJHtEMb

---
_Generated by [Claude Code](https://claude.ai/code/session_01ACY5FuAq6EyXS6tqJHtEMb)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Scrolling Gallery: native border controls** — Width, style, color, and radius replace the old single border-radius field.
 - **SVG Patterns: theme-inheritable default** — A pattern can inherit a "Theme default" preset (type, color, opacity, scale) set at the theme level, so a Style Kit can restyle every pattern site-wide; each block can still override it. (#446)
 - **Row & Grid: Style Kit overlay and hover-state variations** — Row and Grid now support the same background-overlay and hover-state style variations as Section, so a Style Kit's overlay/hover styling applies consistently across all three layout blocks. (#445, #447)
+- **Scroll Slides: overlay opacity control** — The color overlay above each slide's background is now adjustable via an "Overlay Opacity" slider instead of a fixed 80%, so contrast can be tuned when the background image is bright. Existing blocks are unchanged (they keep the 80% default).
+- **Modal: accessible label controls** — New "Accessible Label" and "Close Button Label" fields let authors customize the dialog's and close button's `aria-label` for screen readers. Both default to the previous English strings ("Modal" / "Close modal") when left empty, so existing modals are unchanged.
 
 ### Changed
 - **Icon, Divider, Map, and form field blocks render dynamically** — They always reflect the current theme and store cleaner markup; existing blocks migrate automatically.
@@ -25,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Progress Bar no longer bakes a default color into saved markup** — An unset bar/track color now inherits the theme instead of a fixed hex value. (#440)
 
 ### Bug Fixes
+- **Modal: custom accessible label now works** — The dialog's `aria-label` read a `modalLabel` attribute that was never registered, so a custom label could never take effect and it always fell back to "Modal". The attribute is now registered and wired to the new "Accessible Label" control.
 - **Icon List frontend parity** — Items show their fill / outline and stroke on the frontend, matching the editor.
 - **SVG Patterns / Form Builder color baking** — No longer bake default colors into saved markup, so they inherit the theme's colors.
 - **Abilities API quote / backslash handling** — Content with quotes or backslashes saved through AI-assisted edits is no longer altered.
--- a/src/blocks/modal/block.json
+++ b/src/blocks/modal/block.json
@@ -94,6 +94,10 @@
 			"type": "string",
 			"default": ""
 		},
+		"modalLabel": {
+			"type": "string",
+			"default": ""
+		},
 		"allowHashTrigger": {
 			"type": "boolean",
 			"default": true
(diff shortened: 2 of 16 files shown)


jnealey-godaddy:
Thanks for the review — both findings addressed in c14ab05:

- **Medium (whitespace-only label):** fixed. I guarded the fallbacks with `?.trim()` at each read site in `save.js` and `edit.js` instead of trimming in `onChange`. Trimming on every keystroke would strip a trailing space mid-typing and break entering multi-word labels; guarding at the read site keeps typing natural while still falling back to the default for any blank/whitespace-only value. Since these labels only flow through `save.js`/`edit.js` (no `render.php` on Modal), the four read sites are the complete set.
- **Low (placeholder):** added `placeholder` text ("Modal" / "Close modal") to both label fields so the shipped default is visible in the empty control.

Modal unit suite still green (33/33).

---

jnealey-godaddy:
Both low-severity notes addressed in c734b55:

- **Opacity clamping parity:** `overlayOpacity` is now clamped to `[0, 100]` (and falls back to `80` for non-finite values) in both `scroll-slides/save.js` and `edit.js`, matching `render.php` so the JS and PHP paths agree on out-of-range input.
- **Test coverage:** added save()-output unit tests, since the existing modal suite only exercises the frontend interactivity class:
  - `tests/unit/modal-labels.test.js` — `modalLabel`/`closeButtonLabel` fall back to the default for empty and whitespace-only values, and pass a real value through verbatim.
  - `tests/unit/scroll-slides-overlay-opacity.test.js` — default 80 stays byte-compatible (`0.8`), percent→fraction mapping, zero preserved (not treated as unset), and out-of-range values clamp to `[0, 1]`.

All 12 new tests pass; lint clean.

---

jnealey-godaddy:
DRY note addressed in a8e79d3: extracted the clamp/fallback math into `src/utils/overlay-opacity.js` (`overlayOpacityFraction`), now used by both `scroll-slides/save.js` and `edit.js`, and added a sync-reference comment on the PHP mirror in `render.php`. Output is unchanged. Added `src/utils/test/overlay-opacity.test.js` covering the util directly (boundaries + non-finite fallback). Build, lint, and php -l all clean.

---

jnealey-godaddy:
Thanks for the approval. Both remaining Low notes addressed in f0bd332:

- **Help-text copy:** the fallback also triggers for whitespace-only input, so the `ModalSettings`/`CloseButtonSettings` help now reads "when left blank" instead of "when empty."
- **JS/PHP parity test:** added a data-driven PHPUnit test in `render-test.php` asserting `render.php`'s overlay-opacity formatter matches the JS `overlayOpacityFraction()` util (80→0.8, 50→0.5, 0 preserved, out-of-range clamps to [0,1]). It uses regex boundaries so it doesn't depend on how core serializes the style attribute. This guards the two hand-synced paths against drift, which was the concern behind the "keep in sync manually" comment.

---

jnealey-godaddy:
Both nits handled:

- **`hasValue`/trim mismatch** (b44343b): changed the reset indicator to `label.trim() !== ''` in `ModalSettings` and `CloseButtonSettings`, so a whitespace-only label now reads as the default state in the panel — matching the `?.trim()` fallback that actually renders. Cosmetic, no output change.
- **PHP/JS opacity-format duplication:** leaving as-is intentionally. There's currently a single PHP caller (`render.php`), so extracting a shared helper now would be premature; the sync-reference comment plus the PHPUnit parity test cover the drift risk. Happy to extract `designsetgo_*` helper alongside `designsetgo_safe_css_value()` if/when a second PHP caller appears, as you suggested.

---

PR #450, opened 2026-07-08
Make inline design values (icon gap/size, list gaps, accordion height/gap, marquee width, blob max-width) kit-controllable

jnealey-godaddy:
## Description

Several blocks baked design values as raw inline styles in `save()`, which Style Kits can't theme (inline beats class specificity) and patterns can't hand-edit (any deviation fails WordPress block validation). This PR moves those values out of `save()` into kit-controllable CSS custom-property chains, emitting inline styles only for explicit author overrides — so kits can retheme and patterns can omit the value and inherit.

Each value now resolves through a layered chain: `var(--dsgo-<block>-<prop>, var(--wp--custom--designsetgo--<block>--<prop>, <literal-fallback>))` — a `--dsgo-*` hook a kit's CSS / a semantic class can set, over the theme.json token, over the historical literal. This mirrors the merged shape-divider / svg-pattern theme-inheritance precedent. Every `save()` change ships with a full deprecation (`isEligible` + `save` + `migrate`) so existing content and shipped patterns migrate silently.

**Already shipped, no change needed:** the requested `section` items (overlay via `is-style-overlay-*` variation, and `contentWidth` resolving to `--wp--style--global--content-size`) are already in the codebase (#445), so `section` is untouched.

## Type of Change

- [x] Code refactoring

## Related Issue

Closes #

## Changes Made

- **`icon-button`** — icon↔text gap resolves from `--dsgo-icon-button-gap` → `--wp--custom--designsetgo--icon-button--gap` → `8px` and is omitted entirely when there's no icon (new `dsgo-icon-button--has-icon` marker gates it); icon size gained a `--dsgo-icon-button-size` hook ahead of the existing token. `iconGap` default dropped (unset = inherit); explicit gap still wins inline. Deprecation `v8` + 10 tests.
- **`icon-list-item`** — item gap (12/16px), content gap, and inherited icon-box size moved out of inline into kit-controllable chains keyed on the existing modifier classes; `contentGap` default dropped. Deprecation `v3` + 9 tests.
- **`image-accordion`** — `height` (was `500px`) and `gap` (was `4px`) default to themeable values (`--dsgo-image-accordion-height` over a custom token; `--dsgo-image-accordion-gap` over `--wp--preset--spacing--20`) instead of baked px. Emitted inline only when explicit. Deprecation + 10 tests.
- **`scroll-marquee`** — new `imageWidth: "auto"` (aspect-ratio driven, now the default) via an "Auto width" toggle; the frontend already measures rendered width at runtime, so no JS change. Deprecation `v4` pins existing marquees to their stored width (byte-preserved). + 9 tests.
- **`blobs`** — native, kit-controllable `maxWidth` (`--dsgo-blob-max-width` over a custom token) replacing the generic max-width extension's baked inline value; blobs added to that extension's `EXCLUDED_BLOCKS`. Deprecation `v3` migrates the legacy `dsgoMaxWidth` attribute; the two shipped blob patterns are rewritten to the native form. + 9 tests.

## Testing

- [ ] Tested on frontend
- [ ] Tested responsive behavior (mobile/tablet/desktop)
- [x] No console errors
- [ ] No PHP errors

Verification performed in this environment:
- Full unit suite green — **2277 tests across 81 suites pass**, including new deprecation/save tests for all five blocks that parse pre-change markup and assert silent migration (no "Attempt Recovery").
- `npm run build` compiles cleanly; confirmed each new CSS hook (`--dsgo-icon-button-gap`, `--dsgo-icon-button-size`, `dsgo-icon-button--has-icon`, `--dsgo-icon-list-*`, `--dsgo-image-accordion-*`, `--dsgo-blob-max-width`) is present in the built block CSS.
- `lint-js` and `lint-style` pass on every file changed in this branch. (The repo has pre-existing lint errors in untouched files — visibility extension, a query test — that are out of scope here.)

Live WordPress-editor / frontend / theme / responsive checks were not run in this headless environment.

## Screenshots/Videos

N/A

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have updated documentation as needed (CHANGELOG)
- [ ] I have tested on WordPress 6.4+
- [x] I have followed the patterns in CLAUDE.md
- [x] All files are under 300 lines (if applicable)
- [x] I have added JSDoc comments to new functions
- [ ] Accessibility: WCAG 2.1 AA compliant
- [ ] Security: All user input is validated and sanitized
- [x] Internationalization: All user-facing strings use `__()` or `_e()`

## Additional Notes

Every `save()` change is backed by a deprecation with `isEligible`, so existing posts and the ~100 patterns that emit these inline values migrate silently rather than showing validation warnings — no mass pattern rewrite was required. The `blobs` change also fixes a pre-existing block↔pattern mismatch (patterns emitted a `dsgo-has-max-width` wrapper the block never produced).

https://claude.ai/code/session_01TWS9LADS9NBsaekLjTtYNM

---
_Generated by [Claude Code](https://claude.ai/code/session_01TWS9LADS9NBsaekLjTtYNM)_


jnealey-godaddy:
Thanks for the review — addressed in `dfcc751`.

**HIGH — image-accordion deprecation (fixed).** `migrate()` is now a **passthrough** that pins whatever height/gap the old markup carried, rather than stripping values equal to the old default. Since old markup can't distinguish an explicit `500px`/`4px` from an implicit default, stripping risked silently un-pinning an author's deliberate choice — exactly the "changes unexpectedly when a kit sets the token" concern you raised. Existing accordions now render exactly as authored (same approach scroll-marquee already takes for width); new content/patterns still inherit by omitting the attribute. I also tightened `isEligible` to the old always-both signature (require *both* inline `--dsgo-image-accordion-height:` and `--dsgo-image-accordion-gap:`) and added the regression test you asked for — a current block with an explicit `height:500px` is not flagged eligible and keeps its value.

**MEDIUM — scroll-marquee (no change, intentional).** Its `migrate()` is already a passthrough, so there's no data loss. On the churn concern: WordPress validates a block against the *current* `save()` first and only consults a deprecation's `isEligible` when that fails — so a valid current marquee (even one with an explicit `imageWidth:300px`) matches current `save()` and never reaches `isEligible`. In practice `isEligible` only ever sees genuinely-old markup, so there's no repeated migration/`console.info` on valid content. I left it as-is to avoid unnecessary churn on already-tested code.

**LOW — editor.scss dead fallback (fixed).** Collapsed the nested `var()` chain in `.dsgo-image-accordion__items` to `var(--accordion-gap)` / `var(--accordion-height)`, since style.scss already resolves those on the ancestor.

**LOW — DRY `hasExplicit*` helper (deferred).** Agreed it repeats, but extracting a shared helper now means re-touching save.js **and** edit.js across all five blocks, each of which must stay byte-for-byte identical for block validation. Not worth the parity risk in this PR; better as a focused follow-up.

Full suite still green (2277 tests) and lint clean on all changed files.

---

jnealey-godaddy:
Round 2 addressed in `5ab1fbe`.

**HIGH — blobs v3 `isEligible` nested-block collision (fixed).** You're right: it scanned the whole subtree, so a nested child stamped by the still-active generic max-width extension (same `dsgo-has-max-width` + raw `max-width:`) could false-match a valid native Blobs block. `isEligible` now isolates the **wrapper's own opening tag** — located from the first `dsgo-blobs-wrapper` occurrence (the root class; children never carry it) back to its `<` and forward to its `>`, robust to a leading block-delimiter comment — and tests the signature only there. The old format wrote the raw `max-width` on the wrapper root; the native format writes `--dsgo-blob-max-width` there (excluded by the `[^-]` guard), so a nested child's raw `max-width` no longer lives in the scoped tag. Added a regression test with a native wrapper + a nested `dsgo-has-max-width` child asserting it is not flagged.

**MEDIUM — icon-list-item v3 `isEligible` (fixed).** Same scoping treatment (isolate the item's own opening tag before the `align-items…;gap` test), removing the fragile dependency on nested blocks' inline-style ordering. Regression test added with a nested block inlining `align-items:center;gap:8px`.

**LOW — Content Gap reset (fixed).** Added `allowReset` to the `RangeControl` so a reset clears `contentGap` (`onChange(undefined)`) back to the themeable default instead of pinning a number once touched. (This block uses `PanelBody` rather than the `DsgoInspectorPanel` ToolsPanel the other blocks use, so `allowReset` is the idiomatic in-place affordance here.)

Full suite green (2279 tests, +2 regression) and lint clean on all changed files.

Re the earlier round (already pushed in `dfcc751`): image-accordion `migrate` is now a passthrough (no default-stripping) with a tightened both-props `isEligible` + regression test; the editor.scss dead fallback was collapsed. The DRY `hasExplicit*` extraction I'm still leaving as a follow-up to avoid re-touching the byte-exact save↔edit parity across five blocks in this PR.

---

jnealey-godaddy:
Round 3 addressed in `2ca7cdc`.

**Medium — image-accordion `isEligible` not wrapper-scoped (fixed).** Applied the same opening-tag scoping as blobs/icon-list-item: it now isolates the outer accordion's own opening tag (via `indexOf('dsgo-image-accordion')` → nearest `<…>` bounds) before checking for the inline height+gap props, so a nested `image-accordion` inside an item can't false-trigger the outer, valid block. Regression test added.

**Low — icon-button v8 `isEligible` (fixed).** Same treatment — scoped to the button's own opening tag so free-form button text like "Mind the gap: …" no longer matches the `gap:` signature. Regression test added.

**Medium — migration philosophy consistency (fixed).** Made all deprecations consistent: every `migrate()` is now a **passthrough that pins** whatever the old markup carried. Previously icon-button/icon-list-item stripped a default-valued gap (retroactively themeable) while image-accordion/scroll-marquee pinned. They now all pin — existing content renders exactly as authored (the least-surprising plugin-update behavior), and new content/patterns get the themeable default by omitting the attribute. This removes the ambiguity of un-pinning a value an author may have deliberately set at the old default. Tests updated to assert the pin behavior.

**Low — DRY (fixed).** Extracted `hasExplicitString` / `hasExplicitNumber` into `src/utils/has-explicit-value.js` and wired them into the `save.js`/`edit.js` pairs for icon-button, icon-list-item, image-accordion, and blobs, so the "MUST MATCH" pairs can't drift on that predicate. The `typeof` guards (preserving explicit `0`/`"0px"`) are kept in the helper.

**Nit — icon Gap placeholder.** Documented the static `"8px"` placeholder as a known limitation (won't reflect a kit token override; empty field always means "inherit").

Full suite green (2281 tests, +2 regression) and lint clean on all changed files. Re the checklist: live editor/frontend/theme/responsive passes weren't run in this headless environment — flagging that for a manual smoke test before merge, as the reviewer noted.

---

jnealey-godaddy:
Round 4 addressed in `0a2a614`.

**HIGH — blobs max-width dropped align-aware anchoring (fixed).** Restored in CSS only. The old extension anchored via `has-text-align-*` classes and `align`, and the blobs wrapper already carries `alignleft`/`alignright` (the stylesheet already uses `&.alignright > .dsgo-blobs`), so `.dsgo-has-max-width` now keys its margins off both: `&.alignleft, &.has-text-align-left → margin-left:0; margin-right:auto` and the right-side mirror; otherwise it centers. The extra class outranks the base centering rule, so no `!important` is needed. Mirrored in `editor.scss`. No `save.js`/`edit.js` change is required since WordPress already emits those classes on the wrapper. Added a regression test asserting a left/right-aligned + `maxWidth` blob keeps both its `alignleft`/`alignright` and `dsgo-has-max-width` classes so the anchoring hook applies.

**Medium — duplicated wrapper-tag scoping (fixed).** Extracted `getOwnOpeningTag(innerHTML, rootClassName)` into `src/utils/get-own-opening-tag.js` and wired it into all four deprecations (blobs, icon-button, icon-list-item, image-accordion), removing the four near-identical `indexOf`/`lastIndexOf`/slice copies. Same motivation as the `has-explicit-value.js` extraction.

Full suite green (2282 tests, +1 regression) and lint clean on all changed files; the anchoring rules are present in the built `blobs` CSS.

One note on the anchoring: it's CSS keyed on classes, so it can't be asserted via the jest markup tests directly (those verify class presence, which is the hook). A live editor/frontend check on a left-aligned blob + max-width is worth including in the manual smoke test before merge.

---

jnealey-godaddy:
Round 5 addressed in `bf141e0` — and you're right, which also corrects something I claimed in earlier rounds.

I verified against the WordPress source (`apply-block-deprecated-versions.js`): `if (block.isValid && !isEligible(...)) continue;` — so `isEligible` returning `true` **does** force a migration pass on a block that already validates against the current `save()`. My earlier statements that "valid content validates against current save first and never reaches `isEligible`" were wrong; thanks for the push to check.

Given that, both findings are real:
- **image-accordion v1** — both-props-inline (old default) is byte-identical to a current block that sets both `height` and `gap` explicitly.
- **scroll-marquee v4** — `--dsgo-marquee-image-width:300px` (old default) is byte-identical to a current marquee where the author toggles off "Auto width" and types `300px`.

Fix: **removed `isEligible` from both.** The key realization is that neither needs it — genuinely old content is *invalid* against the current `save()` (which omits the props / emits `auto`), so it reaches the deprecation through the normal `save()`-matching path (`block.isValid === false` short-circuits the `continue` guard regardless of `isEligible`). `isEligible` here was only adding the ability to force-migrate *already-valid* content, which is exactly the false positive. With it gone, valid current content is skipped and only invalid old markup migrates. (The older border-radius-bearing scroll-marquee shapes still route to their own deprecations via `save()`-matching, since v4's `save()` doesn't reproduce that markup.)

The migration tests confirm it: old default markup still migrates and pins its value, and the byte-identical check still passes. I replaced the direct `isEligible(...)` unit tests with parse-based behavioral ones, including new guards that a valid both-explicit accordion / explicit-`300px` marquee is **not** force-migrated (no unexpected `console.info`, which the jest-console matcher enforces).

This also means the icon-button/icon-list-item/blobs `isEligible` checks are the right pattern precisely because their signatures (`--has-icon` presence, item-root `align-items;gap`, raw `max-width:`) are ones the current `save()` can never reproduce — so they don't share this issue. Full suite green (2278 tests), build + lint clean.

---

PR #453, opened 2026-07-10
fix(image-accordion): gate overlay markup on enableOverlay

jnealey-godaddy:
## Description

The Image Accordion block's `save()` unconditionally emitted the three overlay custom properties and a `data-enable-overlay` marker onto the root element — **even when the overlay was disabled**:

```
style="…;--dsgo-image-accordion-overlay-color:#000000;--dsgo-image-accordion-overlay-opacity:0.6;--dsgo-image-accordion-overlay-opacity-expanded:0.35" data-enable-overlay="true"
```

Because the item's overlay pseudo-element is gated on the `--has-overlay` class — which is driven by the `enableOverlay` **context**, not this markup — those values were dead weight whenever the overlay was off. More importantly, a pattern/content author could not remove them from the HTML: deleting the vars triggered a block-validation mismatch because `save()` regenerated them from the still-set attributes (`overlayOpacity`, `overlayOpacityExpanded`, and the `overlayColor` default) on the next parse.

Now the overlay properties and `data-enable-overlay` are emitted **only when `enableOverlay` is true**. Turning the overlay off actually removes it from the saved markup.

## Type of Change

## Related Issue

Closes #

## Changes Made

- `save.js`: gate the three `--dsgo-image-accordion-overlay-*` custom properties and the `data-enable-overlay` marker on `enableOverlay`. The **enabled** output is byte-identical to before, so overlay-on content (the default) is unaffected; a disabled overlay leaves no overlay trace in the HTML.
- `edit.js`: mirror the same gating so the editor preview matches the frontend.
- `deprecated.js`: add a `v2` deprecation reproducing the prior always-emit format, so existing overlay-off content re-serializes into the leaner markup instead of showing WordPress's "unexpected or invalid content / Attempt Recovery" prompt. `migrate` is a passthrough (attributes already describe the block; only the markup changes). No `isEligible` — the enabled output is byte-identical to the current `save()`, so a valid overlay-on block must not be force-migrated.
- `test/deprecated.test.js`: add coverage for the overlay gating (enabled emits props + marker; disabled omits all three props and the marker, ignoring custom overlay attribute values) and the `v2` migration (overlay-on matches current `save()` directly with no deprecation; stale overlay-off markup migrates silently and re-serializes lean).

## Testing

- [x] Unit tests: `wp-scripts test-unit-js src/blocks/image-accordion/test/deprecated.test.js` — 15/15 passing
- [x] `wp-scripts lint-js` on changed files — clean
- [x] `wp-scripts build` — compiles successfully
- [ ] Tested on frontend
- [ ] No console errors

## Additional Notes

Nothing at runtime consumes the parent's overlay custom properties or `data-enable-overlay` attribute when the overlay is off (`data-enable-overlay` is not read by `view.js`; the item's overlay comes from the `enableOverlay` context via the `--has-overlay` class), so omitting them is behavior-preserving for the overlay-off case. The overlay-on output is unchanged.

https://claude.ai/code/session_01VNtw62N423dqmJnvQvNm2S

---
_Generated by [Claude Code](https://claude.ai/code/session_01VNtw62N423dqmJnvQvNm2S)_

### DIFF
--- a/src/blocks/image-accordion/deprecated.js
+++ b/src/blocks/image-accordion/deprecated.js
@@ -193,4 +193,148 @@ const v1 = {
 	},
 };
 
-export default [v1];
+/**
+ * Version 2: Before overlay output was gated on `enableOverlay`
+ *
+ * This format (post height/gap refactor) always baked the three overlay custom
+ * properties (`--dsgo-image-accordion-overlay-color` / `-opacity` /
+ * `-opacity-expanded`) and a `data-enable-overlay` marker onto the root element,
+ * even when the overlay was disabled. Because the item's overlay pseudo-element
+ * is gated on the `--has-overlay` class (driven by the enableOverlay context,
+ * not this markup), those properties were dead weight whenever the overlay was
+ * off — and, more importantly, an author could not remove them from a pattern's
+ * HTML because save() regenerated them from the (still-set) attributes.
+ *
+ * The current save() emits the overlay properties and `data-enable-overlay` ONLY
+ * when `enableOverlay` is true. The enabled output is byte-identical to this
+ * version, so overlay-on blocks (the overwhelming majority, since enableOverlay
+ * defaults true) match the current save() directly and never reach this
+ * deprecation. Only overlay-off blocks — which this version serialized WITH the
+ * now-omitted overlay properties and `data-enable-overlay="false"` — mismatch the
+ * current save() and are routed here so they re-serialize into the leaner format.
+ *
+ * `migrate` is a passthrough: the attributes already describe the block exactly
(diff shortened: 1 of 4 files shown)


jnealey-godaddy:
Superseded by #455, which supersedes this approach.

#455 makes the three overlay attributes nullable and inherits them from CSS (parent var → theme token → literal) instead of gating output on `enableOverlay`. That fixes the overlay-**on** validation case this PR left open (the default is `enableOverlay: true`, so the vars were still baked here), avoids the overlay-off rendering change this PR would have shipped, and keeps `data-enable-overlay` in the markup — #455 uses it as the frontend gate that finally makes 'Enable Overlay: off' remove the scrim on the frontend.

PR #457, opened 2026-07-12
fix: pill default text size + form submissions dropping backslashes

jnealey-godaddy:
Two independent bug fixes found while smoke-testing the 2.4 release, bundled at the author's request. They touch unrelated subsystems (block CSS / form data handling) and can be split if a reviewer prefers.

---

# 1. Pill: default text size regression (`src/blocks/pill/style.scss`)

**A 2.4 regression.** The dynamic-rendering conversion's deprecation drops the old `fontSize: "small"` attribute default so `save()` no longer bakes a `has-small-font-size` class into markup. That part is correct — but **nothing replaced the default**, so the wrapper fell back to its inherited context size.

- **Every pill authored before 2.4 grows from 14px to 22px on upgrade** — silent, across all existing content.
- **Newly inserted pills** render at the inherited size too, visually identical to a pill explicitly set to `large`.

The asymmetry is the tell: the *same* deprecation also drops the baked `align: "center"` default, but there a CSS rule (`.dsgo-pill:not(.alignleft):not(.alignright) { justify-content: center }`) reproduces the old default. Font size had no counterpart.

**Fix** — supply the default in CSS at zero specificity, so markup stays clean and themeable while every explicit choice still wins:

```scss
:where(.dsgo-pill) {
	font-size: var(--wp--preset--font-size--small, 0.875rem);
}
```

A **preset** adds `has-*-font-size` to the wrapper — (0,1,0) beats `:where()`'s (0,0,0). A **custom** size is emitted against `.dsgo-pill__content` by the fontSize support's `__experimentalSelector`. The inner span declares no font-size of its own, so it inherits whichever applies.

**Verified** against content authored on a real `v2.3.0` build, then upgraded to this branch:

| Case | Before | After |
|---|---|---|
| Legacy pill, default (migrated) | 22px ❌ | **14px** ✅ (matches 2.3 baseline) |
| Legacy pill, explicit `large` | 22px | **22px** ✅ |
| Fresh pill, no size set | 22px ❌ | **14px** ✅ |
| Fresh pill, explicit `small` | 14px | **14px** ✅ |
| Fresh pill, custom `30px` | 30px | **30px** ✅ |

Editor and frontend agree in every case.

---

# 2. Forms: submissions silently dropped backslashes (`includes/blocks/forms/class-form-handler.php`)

**Not a 2.4 regression — this predates the release** (the form handler is untouched in 2.4 apart from a phpcs pass). Found while verifying the form-field blocks still worked after their dynamic conversion.

`update_metadata()` runs `wp_unslash()` on every value it stores, so it expects **slashed** input. `store_submission()` was handing it values that had *already* been unslashed (REST params arrive unslashed; the no-JS `$_POST` path unslashes explicitly). WordPress then removed a second level of escaping that was never there.

**Result: any backslash in a submitted value was silently eaten.**

```
submitted: C:\Users\me\report.txt
stored:    C:Usersmereport.txt
```

Windows paths, code snippets, regexes, and LaTeX pasted into a message field were all corrupted. This is the **same defect class already fixed for the Abilities API in `b41ab58d`** (`wp_slash()` before `wp_update_post()`); it had simply never been applied to form submissions.

**Fix** — re-slash the user-controlled values before storing. `wp_slash()` walks arrays, so the whole `$fields` structure is covered; user agent and referer get the same treatment as user-controlled strings stored the same way.

**Tests** — two regression tests added to the existing `Test_Form_Handler` suite. I confirmed the first one **fails on the prior code with exactly the reported corruption** (`C:\Users\me\report.txt` → `C:Usersmereport.txt`), so it's a real guard rather than a passing no-op. The second guards the other direction: that re-slashing doesn't leave stray escape characters behind in quotes or unicode.

Also verified end-to-end in a browser — submitting a Windows path and a regex through a real contact form now stores both intact.

---

## Verification

- `npm run build` clean; `npx eslint` / `stylelint` clean on touched files (the repo's pre-existing lint errors are unchanged)
- **PHPUnit `Test_Form_Handler`: 37/37 passing**, including the 2 new tests
- Pill Jest suite: 5/5 passing
- e2e suite: 7/7 passing
- Contact form re-verified in browser: the shipped `designsetgo/contact/contact-form` pattern inserts with zero invalid blocks, renders, validates, submits, and stores all values correctly

## Changelog

Also removes a stale `CHANGELOG.md` entry that described the pill regression as *intended* behavior ("Set an explicit font size on any pill that should keep the smaller look") — it would have shipped as guidance — and adds entries for the form fix to both `CHANGELOG.md` and `readme.txt`.


jnealey-godaddy:
Thanks — went through both notes. Actioned the first, and I'm pushing back on the second with evidence.

## Medium (changelog) — valid, fixed in 8953ffd

Confirmed and it's a good catch. `CHANGELOG.md` line 23 didn't just contradict the fix, it actively instructed authors to work around the bug ("Set an explicit font size on any pill that should keep the smaller look"). That would have shipped as guidance.

Removed it and folded the correction into the existing "Pill renders dynamically" bullet, which already said the centered alignment is CSS-driven and unchanged — the small text size now sits in exactly the same position, so the two read consistently.

One correction to the finding: **`readme.txt` did not contain that bullet.** Its only relevant line is "saved pills no longer bake in a fixed size or alignment", which remains accurate after this fix (the size comes from CSS rather than a baked `has-small-font-size` class), so it needs no change. Only `CHANGELOG.md` carried the stale entry.

## Low (editor parity) — not applicable here; no change made

The "must duplicate into `editor.scss`" convention doesn't apply to this rule, because for the pill both stylesheets already compile into a **single shared bundle**:

- [`src/blocks/pill/index.js:16-17`](src/blocks/pill/index.js#L16-L17) imports **both** `./editor.scss` **and** `./style.scss`
- `block.json` points **both** `"style"` and `"editorStyle"` at that same `file:./index.css`
- Verified in the build output: `grep ":where(.dsgo-pill)" build/blocks/pill/index.css` → present (1 match). The rule is already in the editor bundle.

So duplicating it into `editor.scss` would emit the identical declaration twice into the same compiled file — redundant, and one more place to drift. The existing "CRITICAL: must duplicate" comments in that file look like a leftover from an earlier build setup rather than a live constraint.

Empirically confirmed too, which is what prompted me to check: measuring the pill inside the editor canvas iframe gives **14px** (default), **14px** (explicit `small`), **~29.3px** (custom `30px`, fluid clamp) — matching the frontend exactly in all three cases. The `:where()` zero-specificity concern is the right thing to worry about in general, but nothing in the editor chrome sets `font-size` on `.dsgo-pill`, and the measured values confirm it.

Happy to reconsider if the duplication convention exists for a reason I'm not seeing.

PR #459, opened 2026-07-12
refactor: move constant layout declarations out of saved markup into CSS

jnealey-godaddy:
## What

Five static blocks serialized CSS declarations into every saved block that **never varied by attribute**. They bloated the database and every pattern file, and — because an inline style beats any stylesheet rule — a Style Kit could never retheme them.

The trigger: Icon Button emitted this on every single button, for a block with no size set.

```html
<span class="dsgo-icon-button__icon dsgo-lazy-icon"
      style="display:flex;align-items:center;justify-content:center;flex-shrink:0"
      data-icon-name="lightbulb"></span>
```

None of those four declarations depend on an attribute. The proof they were never needed: sibling block **Modal Trigger** already declared them in `style.scss` and emitted no `style` attribute at all. Icon Button was the outlier.

| Block | Baked into `save()` | Status of the CSS |
|---|---|---|
| `icon-button` | `display / align-items / justify-content / flex-shrink` on the icon | rule added |
| `icon-list-item` | same three (CSS already had `flex-shrink`) | rule added |
| `scroll-accordion` | `width / align-self` + `display / flex-direction` | **already in CSS, verbatim** |
| `flip-card` | `width:100%` | **already in CSS** |
| `countdown-timer` | `display:none` on the completion message | **already in CSS** |

Three of the five were pure duplication — the stylesheet already said the same thing, so the inline copies did nothing except make the values unoverridable.

Attribute-driven values (colour, padding, gap, an explicit `iconSize`, `scroll-accordion`'s `alignItems`, `flip-card`'s `--dsgo-flip-duration`) stay inline. Only the constants moved.

## Also fixed

**`class-block-inserter.php` (Abilities API), Icon Button branch** — a pre-existing bug found on the way through. It unconditionally emitted `width:20px` + `data-icon-size="20"` even when the caller set no `iconSize` — markup `save()` would never produce, so AI-inserted buttons failed block validation. Same class as the Modal Trigger bug fixed in #458.

## Migration

New deprecations reproduce the old markup exactly; every `migrate()` is a passthrough (markup-only change). `scroll-accordion` and `flip-card` had no `deprecated.js` at all — created and wired into their `index.js`.

**96 pattern files regenerated** via `tools/regenerate-patterns.js`. Net effect on `patterns/`: **995 deletions vs 361 insertions.**

Verified in a real editor + frontend (wp-env, Twenty Twenty-Five): every element takes its layout from CSS with computed values identical to before, and pre-change stored markup migrates silently — `isValid: true`, no "Attempt Recovery", attributes intact.

## Two landmines worth reviewing

**Deprecations were being selected positionally.** Both `icon-button`'s and `icon-list-item`'s tests did `const [v9] = deprecated` — prepending a newer entry silently re-points that at the wrong version. It surfaced as 7 then 4 red tests. All deprecation entries are now **exported by name**.

**A test fixture that was quietly self-invalidating.** `icon-list-item`'s test built its "old markup" by string-patching the *current* `save()` output, anchored on `justify-content:center`. The moment that declaration was removed, the patch stopped matching and the fixture silently became a no-op — it would have kept "passing" against nothing. It now derives from `v4.save()`, a frozen reproduction that can't drift.

Also documented: WordPress calls `isEligible(attributes, innerBlocks, { blockNode, block })` — there is **no `innerHTML` key** on that third argument, so the codebase's existing `({ innerHTML })` destructures to `undefined`. Those checks are dead; they work anyway because for an *invalid* block WordPress skips `isEligible` and matches on `save()` output. The new deprecations use the real signature.

## Deliberately out of scope

`grid` (`display:grid`, 101 patterns) and `row` (`display:flex`, 86 patterns) have the same anti-pattern, but they are **structural** — if the stylesheet fails to load, the layout doesn't degrade, it collapses. Different risk profile, ~190 pattern files, and it deserves its own PR.

The shared `dsgo-justify--left` class is also dead (`_justification.scss` has no `&--left` rule; left is the implicit `flex-start` default), but removing it means freezing a local copy of the class map inside each `deprecated.js` first — the deprecations import `getJustificationClass()`, so changing that helper would retroactively alter what the old `save()` functions emit and break their byte-reproduction. Follow-up.

## Verification

- Jest **2311/2311**
- PHPUnit **845/845**
- PHPStan clean (needs `--memory-limit=1G`; the default 128M crashes the worker)
- Editor + frontend checked in wp-env, incl. legacy-content migration

`lint:css` reports one **pre-existing** failure in `src/blocks/slider/editor.scss:275` (duplicate selector) that reproduces on `main` and is untouched here.


jnealey-godaddy:
Both Critical findings were correct. Fixed in `7435d142`.

## 1. Pattern regeneration deleted authored content — confirmed and repaired

You were right, and the diagnosis was exact. I verified the damage before fixing: **16 files, 269 block comments destroyed.**

```
features/features-flip-card-grid.php     33 -> 9   (lost 24)
team/team-flip-cards.php                 35 -> 8   (lost 27)
pricing/pricing-tabs.php                125 -> 97  (lost 28)
content/icon-features.php                22 -> 12  (lost 10)
...
```

Root cause is as you described: `regeneratePatterns()` registered only the target block, so nested `core/heading` / `core/paragraph` / `designsetgo/flip-card-face` were unregistered types at parse time and the parser dropped them.

Worth naming *why* it slipped through: Icon Button is a **leaf** block — its label is an attribute, not `innerBlocks` — so the first batch of 75 files really was a clean CSS-constant removal, and I carried that assumption into the second batch without re-checking. The verification I ran (leftover inline styles, block validity, editor/frontend rendering) all passed on emptied containers.

**Fix:** register a passthrough handler for unregistered types. The parser then routes unknown blocks through `createMissingBlockType()`, which preserves their full delimited markup — nested children included — and `serialize()` emits it verbatim, so inner blocks round-trip byte-for-byte while the target still re-renders through its real `save()`.

One trap while implementing it: **every** support on the passthrough must be off. With `className` enabled, `RawHTML` receives a prop and renders a real `<div>`, injecting `<div class="wp-block-designsetgo-regenerate-passthrough">` around the preserved content. The block-count guard did **not** catch that — a wrapper div adds no block comment — so it needed its own check.

**Two guards, both tested** (`tests/unit/tools/regenerate-patterns.test.js`):
- `assertNoContentLoss()` — throws if any block-name count changes in a region.
- `assertNoPassthroughLeak()` — throws if the scaffold reaches the output.

One test deliberately removes the handler and asserts the guard *throws*, so it can't rot into a no-op.

**Verification after re-regenerating everything from `main`:**
- 0 files with block-count drift
- 0 passthrough leaks
- visible text of all 88 changed files **byte-identical to main** (tags/attrs stripped, hashed, compared per file)
- parsing all 95 shipped patterns in the editor: **3,229 blocks, 0 invalid**

## 2. Four stale Abilities-API branches — fixed

All four (`icon-list-item`, `flip-card`, `scroll-accordion`, `countdown-timer`) now reproduce their block's `save()` output.

`icon-list-item` was worse than described: besides the icon layout, it also baked an **item gap** and a **fixed 32px icon size** that `save()` never emits — `save()` receives no block context, so the parent Icon List's `iconSize` resolves to the `--inherit-size` path. I checked the generator's real output against the ground-truth pattern markup, and confirmed all four parse `isValid: true`.

## Suites

Jest 2317/2317 · PHPUnit 845/845 · PHPStan clean · new tool files lint clean (the 38 `src/blocks` lint errors are pre-existing on `main`).

Thanks — the content-loss catch was the one that mattered, and my own checks would not have found it.

jnealey-godaddy:
Both Low notes addressed in `ae812b92` — and verifying them turned up a **pre-existing validation bug** in the same file that neither of us had flagged.

## The bug (found while re-validating)

`block.json` defaults `linkTarget` to `_self` for both `icon-button` and `icon-list-item`, so a parsed block always carries it and `save()` always emits `target="_self"` alongside `href`. Both Abilities branches defaulted it to `''` and only emitted the attribute when truthy:

```php
$link_target = isset( $attributes['linkTarget'] ) ? $attributes['linkTarget'] : '';
...
if ( $link_target ) { $extra_attrs .= ' target="…"'; }
```

So **every AI-inserted *linked* icon button or icon list item** produced markup `save()` never generates and failed validation ("Attempt Recovery") on first open. Same bug class as the four branches already fixed here — it just wasn't in a code path this PR had otherwise touched.

Caught by round-tripping the generator's real output through `parse()` rather than eyeballing it. First divergence:

```
expected:  … href="#" target="_self"><span class="dsgo-icon-button__icon" …
generated: … href="#"><span class="dsgo-icon-button__icon" …
```

All four variants now validate (`isValid: true`): outlined+icon, `_blank`+rel, unlinked, and a linked list item.

## Your two notes

1. **`:1045` intermediate variables** — right, they were constants after the context fix. Inlined, keeping the one-line note about *why* they're literals (they're save.js's empty-context values, not a choice made in this branch).

2. **`:1093` iconStyle / strokeWidth validation** — done. `iconStyle` is now checked against block.json's enum and `strokeWidth` must be numeric. Agreed it was never an XSS issue (output was already `esc_attr()`-escaped); the real risks were an out-of-enum `iconStyle` emitting a `data-icon-style` the frontend injector doesn't understand, and a non-scalar `strokeWidth` stringifying to `"Array"` with a warning. Confirmed both hostile inputs are now dropped cleanly.

Suites: Jest 2317/2317 · PHPUnit 845/845 · PHPStan clean · PHPCS clean.

PR #461, opened 2026-07-12
fix: version block assets with the plugin version so they cache-bust on release

jnealey-godaddy:
## The bug

Found on a live Airo-generated site running 2.4: Icon Buttons rendered as **giant red squares** with the arrow blown up to fill the button and the label pushed underneath. It cleared after a few hard refreshes — which is the tell.

## Root cause

WordPress versions a block's own CSS/JS with the `version` field from its `block.json` ([`wp-includes/blocks.php`](https://github.com/WordPress/WordPress/blob/master/wp-includes/blocks.php)):

```php
$block_version = ! $is_core_block && isset( $metadata['version'] ) ? $metadata['version'] : false;
$version       = $style_path_norm && SCRIPT_DEBUG ? filemtime( $style_path_norm ) : $block_version;
```

**63 of our 66 blocks still carry the scaffolded `"version": "1.0.0"`.** It has never been bumped. So `build/blocks/{block}/index.css?ver=1.0.0` is byte-identical across every release — browsers and CDNs keep serving the copy they cached from an older version of the plugin, indefinitely.

The live site proves it:
```
build/blocks/icon-button/index.css?ver=1.0.0   <- block.json version, never changes
build/style-index.css?ver=2.4.0                <- plugin version, busts correctly
```

That global stylesheet busting correctly is exactly what masked the problem.

**Look at the `SCRIPT_DEBUG` branch.** In development the version is a `filemtime()`, so block assets always bust and the bug is *impossible to observe locally*. It only exists in production.

## Why 2.4 made it visible

2.4 moved Icon Button's icon layout (`display:flex; align-items; justify-content; flex-shrink`) out of the saved markup and into the stylesheet. That's a good change — but it means the markup now **depends** on the new CSS arriving. With a stale cached stylesheet, the icon span loses its box, and its SVG (`width: 100%`) expands to fill the button.

The latent bug has been there the whole time. Any CSS-only fix to any of the 66 blocks has been silently failing to reach existing users. 2.4 just made the consequence loud.

## The fix

`Blocks\Loader::sync_asset_version()` forces every `designsetgo/*` block's asset version to `DESIGNSETGO_VERSION` via the existing `block_type_metadata` filter. One hook, all 66 blocks, and no more remembering to bump `block.json` on every release. Core and third-party blocks are untouched.

## Tests

`tests/phpunit/block-asset-version-test.php` (6 tests).

The real-world guard feeds **every real `block.json` on disk** through the filter rather than reading the registered style handles — because the test suite runs with `SCRIPT_DEBUG` on, so asserting on the handles would pass with or without the fix. That would have been a vacuous test.

Verified the tests have teeth: removing the filter produces **3 failures + 1 error**.

## Verification

PHPUnit 851/851 · PHPStan clean · PHPCS clean · Jest 2317/2317.

## Note on the other console errors from that site

The `form-builder` and `grid` block-validation failures in the same console are **unrelated to this**. Their stored markup is indented and carries hand-rolled values (`minmax(480px, 1fr)`, inline `--dsgo-form-field-spacing` vars) that our `save()` never emits — Airo is authoring DesignSetGo block markup directly instead of going through block attributes. That's an integration-contract issue on their side, not a caching one.


jnealey-godaddy:
Both findings were correct. Fixed in `0f9c305a`.

**Medium — docblock contradicted the code.** You're right, and it was the more useful of the two: the comment claimed a carve-out ("Blocks that deliberately pin their own version keep it") that doesn't exist in the code, and `scroll-marquee` really does pin `1.2.0` and really does have it discarded.

The unconditional override is intended, so the comment now says that plainly and explains *why* — a per-block pin only busts when someone remembers to bump it, which is precisely the failure mode this fix exists to remove. Preserving pins would reintroduce the bug for any block that has one.

**Low — `@since 2.4.1`.** Corrected to `2.4.0`; `DESIGNSETGO_VERSION` is `2.4.0` and the CHANGELOG entry is under `[2.4.0] - Unreleased`.

**One thing I tightened beyond the note.** The pinned-version test used a *fabricated* `icon-list-item` / `1.2.0` fixture — which is how the docblock and the test managed to disagree with reality without anything failing. It now reads `scroll-marquee`'s real `block.json`, and first asserts that it still pins a version of its own:

```php
$this->assertNotSame(
    DESIGNSETGO_VERSION,
    $metadata['version'],
    'This test is only meaningful while scroll-marquee pins a version of its own.'
);
```

So if that pin is ever removed, the test says so rather than quietly passing against nothing.

Re-verified the suite still has teeth: 6 pass with the filter, **3 failures + 1 error** without it.

PHPUnit 851/851 · PHPStan clean · PHPCS clean.

PR #462, opened 2026-07-13
fix(deprecations): give isEligible the contract WordPress actually calls it with

jnealey-godaddy:
## What

Every one of our 67 deprecation guards was dead code, and switching them on found two blocks losing attributes on `main` today.

WordPress calls a deprecation's `isEligible` as:

```js
isEligible( attributes, innerBlocks, { blockNode, block } )
```

(`@wordpress/blocks` → `api/parser/apply-block-deprecated-versions.js`). All 67 of ours destructured **`{ innerHTML }`**, which is not a key on that object. Each guard opens with a truthiness check, so every one returned `false` unconditionally.

## Why nobody noticed

`isEligible` only matters for a block that is still **valid**:

```js
if ( block.isValid && ! isEligible( ... ) ) continue;
```

For an **invalid** block — the normal case, since these deprecations all exist because markup changed — `block.isValid &&` short-circuits, `isEligible` is never called, and WordPress instead picks the version whose `save()` reproduces the stored HTML. That path carried every migration we had, so the broken guards were invisible.

## What switching them on exposed

Reading the markup from `blockNode.innerHTML` (falling back to `block.originalContent`) makes the guards evaluate for real. Six then claimed **current** content, and two were actively destructive:

| Block | Guard | Effect |
|---|---|---|
| `tab` v1 | `typeof attributes.iconPosition === 'undefined'` | migrated current tabs through a pre-`strokeWidth` schema → **`strokeWidth` silently dropped** |
| `blobs` v2 | `innerHTML.includes('dsgo-blobs-wrapper') && align === undefined` | migrated current blobs through a pre-`height` schema → **`height` silently dropped** |
| `grid` v1, `row` v1, `section` v1 | `attributes.align === undefined` | claimed nearly every current block |
| `form-builder` v4 | `!attributes.fieldBorderColor` | claimed every form with a default border colour |

The shared root cause for four of them: **"attribute absent from the block comment" does not mean "old block."** `attributes` is the raw comment JSON, and WordPress never serializes an attribute equal to its default — so `align === undefined` is just as true of a brand-new block.

`tab` and `blobs` are **pre-existing** bugs (their guards are attribute-only and never needed `innerHTML`); they are live on `main`. The invariant test below fails on `main` for exactly that reason.

All six are markup-change deprecations, which are reached by save-matching on an invalid block and never consult `isEligible` at all — so the guards were pure downside. Removed, each with a comment explaining why.

## Guard against regressions

`tests/unit/deprecations-isEligible.test.js` pins the invariant for every block with deprecations:

1. its own current `save()` output parses valid;
2. **no deprecation claims it**;
3. `createBlock → serialize → parse` round-trips without losing an attribute.

Run against the pre-fix tree it produces 6 failures, including the two attribute losses.

## Also

- Block test call sites passed `{ innerHTML }` as well, so they were asserting against a contract WordPress does not use. Updated to `{ blockNode: { innerHTML } }` — they now exercise real behaviour.
- `CLAUDE.md` taught the wrong mechanics ("Without `isEligible`, users see 'Attempt Recovery'"), which is what produced this whole class of bug. Rewritten to describe the actual algorithm, plus the two traps.

## Verification

- `npx jest` — **90 suites / 2441 tests pass** (baseline on `main`: 89 / 2332; `jest.config.js` untouched).
- `npm run build` — compiles clean.
- `npm run lint:js` — no new errors vs. `main`.
- Pre-commit e2e (7 tests) pass.

No behaviour change for invalid blocks: that path never called `isEligible` and still doesn't. The only semantic change is that valid blocks are no longer spuriously migrated.

### DIFF
--- a/.claude/claude.md
+++ b/.claude/claude.md
@@ -148,35 +148,46 @@ See [`docs/plans/2026-04-17-theme-3-inspector-ia.md`](../docs/plans/2026-04-17-t
 
 Required when changing: attribute schema, HTML structure, or removing attributes.
 
-**Every deprecation MUST have all three**: `isEligible`, `save`, and `migrate`.
+**`save()` is what makes migration silent — not `isEligible`.** WordPress applies a deprecation like this (`@wordpress/blocks` → `api/parser/apply-block-deprecated-versions.js`):
 
-- `isEligible(attributes, innerBlocks, { innerHTML })` — Lets WordPress skip save validation and go straight to migration (silent, no warning). Use attribute checks or innerHTML pattern matching to identify the old version.
-- `save()` — The old save function (best effort reproduction of old output).
-- `migrate(attributes, innerBlocks)` — Transforms old attributes to new format. Use passthrough `return attributes;` if only HTML structure changed.
+```js
+const { isEligible = stubFalse } = deprecatedDefinitions[i];
+if ( block.isValid && ! isEligible( parsedAttributes, block.innerBlocks, { blockNode, block } ) ) continue;
+// ...then: does THIS version's save() reproduce the stored HTML? If not, skip it.
+```
 
-Without `isEligible`, users see "Unexpected or invalid content" warnings with an "Attempt Recovery" button instead of silent auto-migration.
+Read that `block.isValid &&` carefully — it drives everything:
+
+- **The block is INVALID** (the normal case: markup changed, so stored HTML no longer matches the current `save()`). `block.isValid` is `false`, the condition short-circuits, and **`isEligible` is never called**. WordPress picks the version whose `save()` reproduces the stored HTML. Silent, no "Attempt Recovery". A markup-change deprecation therefore needs **no `isEligible` at all** — and adding one buys nothing.
+- **The block is VALID.** `isEligible` is the *only* thing consulted, and returning `true` opts an otherwise-fine block into a re-migration. This is the *only* reason `isEligible` exists: an attribute-only migration where markup did not change.
+
+So: **`isEligible` is a switch for migrating VALID blocks. It cannot rescue invalid ones, and it never suppresses an "Attempt Recovery" warning — only a `save()` that reproduces the stored HTML does that.**
+
+Two traps that follow, both of which have bitten us:
+
+1. **The third argument is `{ blockNode, block }` — there is no `innerHTML` key.** Reach the markup via `blockNode.innerHTML` (or `block.originalContent`). Destructuring `{ innerHTML }` yields `undefined`, so a guard like `innerHTML && innerHTML.includes(...)` silently returns `false` forever. It *looks* like it works, because the invalid-block path above carries the migration regardless.
+
+2. **"Attribute absent from the comment" does NOT mean "old block".** `attributes` here is the raw comment JSON, and WordPress never serializes an attribute whose value equals its default. `attributes.align === undefined` or `typeof attributes.iconPosition === 'undefined'` is therefore just as true of a brand-new block that left the value alone. Guards like that claim *current* content, and then `migrate()` runs it through a schema that predates the block's newer attributes — silently dropping them.
 
 ```javascript
 const v1 = {
-	attributes: { /* old attribute schema */ },
-	isEligible(attributes, innerBlocks, { innerHTML }) {
-		// Identify old blocks by attribute signature or HTML patterns
-		return !Object.prototype.hasOwnProperty.call(attributes, 'newAttribute');
-		// or: return innerHTML && !innerHTML.includes('new-class');
-	},
-	save({ attributes }) { /* old save output */ },
-	migrate(attributes) {
+	attributes: { /* the FULL attribute schema this version had */ },
+	supports: { /* the FULL support set this version had */ },
+	save({ attributes }) { /* reproduce the old output byte-for-byte */ },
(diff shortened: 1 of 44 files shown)


jnealey-godaddy:
Addressed the DRY finding in `2eb12246`.

`getDeprecatedBlockHTML( extra )` ([`src/utils/deprecated-block-html.js`](https://github.com/jnealey-godaddy/designsetgo/blob/2eb12246/src/utils/deprecated-block-html.js)) now owns the expression; all **71 call sites across 32 files** call it instead of re-deriving it:

```js
isEligible( attributes, innerBlocks, extra ) {
    const innerHTML = getDeprecatedBlockHTML( extra );
    ...
}
```

Worth doing beyond the usual DRY argument: the bug this PR fixes *was* 67 copies of a subtly wrong version of this same expression. A single definition is what stops that recurring — which is also why the helper got its own test ([`src/utils/test/deprecated-block-html.test.js`](https://github.com/jnealey-godaddy/designsetgo/blob/2eb12246/src/utils/test/deprecated-block-html.test.js)) rather than relying on the guards' coverage. It pins the contract directly, including that a legacy `{ innerHTML }` shape yields `''` rather than matching, so a future change to the third argument fails one obvious test instead of silently neutering every guard again.

The helper returns `''` rather than `undefined`, so callers keep using `.includes()` / regex tests unguarded and a missing third argument can't produce a false positive.

Verification: **91 suites / 2448 tests pass**, `npm run build` clean, no new `lint:js` errors vs `main`, pre-commit e2e (7 tests) pass.

jnealey-godaddy:
Good catch — I took this one rather than deferring it, because the gap turned out to be load-bearing. Fixed in `48795085`.

You were right that the invariant only ran against `createBlock(name, {})`. The sweep now probes **each attribute in isolation with a non-default value** (enums walk their members, booleans flip, numbers bump, strings take a length; `align` is driven off `supports.align`) — **~479 variants across 36 blocks**, up from 36.

It found **seven more guards claiming current content**, none of which the defaults-only test could see:

**Narrowed** — the markup signature is real, but the *current* `save()` emits it too, so each needed the conjunct that actually separates the eras:

| Guard | Why it over-claimed | Fix |
|---|---|---|
| `grid` legacyMinWidth | a `minmax()` track is also what the current `save()` emits once `columnMinWidth` is set | require `!attributes.columnMinWidth` — legacy content has the track baked into the HTML with no attribute, which is precisely what its `migrate()` recovers |
| `form-builder` v5 | old `save()` wrote `--dsgo-form-input-height` unconditionally; the current one only when `inputHeight` is set | require `!attributes.inputHeight` |
| `icon-button` v7 | the pre-token era baked the inline `width/height` pair in even at the *implicit* default size; the current `save()` writes it only for an explicit `iconSize` | require `iconSize === undefined` |

**Removed** — markup-change deprecations with no signature separating old from current, so `isEligible` could only do harm (it's skipped outright for the invalid blocks these actually serve):

- `row` v3 — fired on any current row with a non-default `tagName`.
- `row` v2 and `section` v2 — keyed on an attribute being **present** in the comment, which is true of any current block whose value is non-default. That's the mirror image of the absence trap this PR already fixes, and I'd missed it.
- `section` v5/v4/v3 — all three shared one guard, `!!(shapeDividerTop || shapeDividerBottom)`, equally true of a current section. A single divider claimed all three versions at once.

Worth stating plainly: **none of these lost attributes** — the round-trip leg stayed green throughout — so the harm was a needless re-migration of valid content, not data loss like `tab`/`blobs`. But the invariant is the point, and it now holds across the attribute space rather than at one point in it.

Verified the sweep has teeth: reverting the `grid` narrowing alone turns it red (`deprecated[0] claims columnMinWidth="7px"`).

**91 suites / 2448 tests pass**, build clean, no new `lint:js` errors vs `main`, pre-commit e2e (7 tests) pass.

jnealey-godaddy:
Correct, and a genuine miss on my part — fixed in `01048bc4`.

My earlier sweep only rewrote `src/blocks/*/test/*.js` and never touched `tests/unit/`, so this call site kept passing the old `{ innerHTML }` shape — the very shape this PR establishes WordPress never sends. The guard therefore read `''` and returned `false` regardless of the markup handed to it, so `toBe(false)` passed for the wrong reason. Exactly the failure mode the PR is about, reproduced inside the PR's own test suite.

It now passes `{ blockNode: { innerHTML } }`. `v4` does genuinely opt out of the legacy nested-div markup — its guard is "block root is not a `<div>`", and that fixture starts with one — so the assertion still holds, just for the right reason now.

I also added the positive counterpart through the same call, since that's what would have caught this in the first place:

```js
expect( v4.isEligible( {}, [], { blockNode: { innerHTML } } ) ).toBe( false );

// …and that `false` is a real answer, not an always-false guard:
expect( v4.isEligible( {}, [], {
    blockNode: { innerHTML: '<button class="dsgo-modal-trigger wp-block-button__link">Open</button>' },
} ) ).toBe( true );
```

A `false` expectation on its own cannot distinguish an opted-out guard from an always-false one — which is precisely how this survived.

Swept the whole repo (`tests/`, `src/`, `tools/`) to confirm no other `isEligible` call site still uses the old shape. This was the last one.

**91 suites / 2448 tests pass**, no new `lint:js` errors vs `main`, pre-commit e2e (7 tests) pass.

PR #463, opened 2026-07-13
fix(security): close CSS escape bypass, escape form emails and block classes

jnealey-godaddy:
Found by the pre-release audit of `v2.3.0..main` while staging 2.4.0.

**All four are pre-existing** — present verbatim in the shipped `v2.3.0` tag, so none is a 2.4.0 regression and shipping 2.4.0 would not have made anything worse. But we know about them now, so they get fixed before the next release rather than after it.

## 1. CSS escape-sequence bypass

`includes/features/class-style-binding.php`, `includes/abilities/class-css-sanitizer.php`

A browser resolves CSS escapes **before** applying a declaration. So `\75\72\6c(//evil.test)` *is* `url(//evil.test)`, and `java\0script:` *is* `javascript:`. Both sanitizers pattern-matched the **raw** text:

```php
if ( preg_match( '/url\s*\(|expression\s*\(|javascript:|data:/i', $value ) ) { continue; }
```

…which blocks nothing. The attacker escapes one character and walks through.

The instructive part: `Custom_CSS_Renderer` **already did this correctly**, and that is exactly why the other two didn't — the normalization was a `private` method, so it couldn't be reused. It now lives in `designsetgo_normalize_css_escapes()` (`includes/helpers.php`) and all three call it. One definition of "what does the browser actually see."

`StyleBinding` additionally rejects any value whose decoded form differs from its raw form. Bindings resolve to lengths, colours and keywords — none has any reason to be escaped — so rather than trust this decoder to agree with every browser tokenizer on every edge case, the value is dropped.

## 2. HTML injection into notification emails

`includes/blocks/forms/class-form-handler.php`

Notifications are sent `Content-Type: text/html`, and the per-field merge tags were interpolated **raw**:

```php
$merge_tags[ '{' . $field_name . '}' ] = is_array( $field_data ) ? $field_data['value'] : $field_data;
```

The `{all_fields}` aggregate three lines below always ran values through `esc_html()`. So any template using `{message}` or `{name}` — the common case — let an **anonymous** submitter put arbitrary markup into the site owner's inbox: a fake "reset your password" link, a tracking pixel, a spoofed message.

Every merge tag is now escaped. Array values (checkbox groups, multi-selects) are joined rather than handed to `esc_html()`, which would have raised on them.

## 3. Attribute breakout in generated core-block HTML

`includes/abilities/class-block-inserter.php`

`textAlign` / `align` were concatenated into `class="..."`, so `center" onmouseover="alert(1)` closed the attribute and added a live handler. These are alignment keywords, so they're now constrained to the set core actually emits — **provably safe rather than merely escaped**, and dropping a nonsense alignment is also the correct rendering.

Reaching this needs the `add-block` ability (`edit_posts`) *and* `unfiltered_html` for the payload to survive `wp_update_post()`'s KSES pass — so it is not an escalation past an existing trust boundary. It's still a real hole.

## 4. Tabs deep-link crash

`src/blocks/tabs/view.js`

`window.location.hash` went straight into `querySelector('#' + hash)`. A fragment is **not** a valid CSS selector in general: `#2024`, `#a b`, `#!` all throw `SyntaxError` — thrown from inside `init()`, which took out **every tabs block on the page**, not just the deep link. An ordinary year-numbered anchor elsewhere on the page was enough. Now `CSS.escape()`d, with a throw treated as "no match" rather than fatal.

## Tests

28 new PHPUnit assertions across three files, plus 7 jsdom tests for the fragment handling. **Each was verified to fail against the unpatched code** — 6 PHPUnit failures, and `SyntaxError: not a valid selector` in jsdom. They're regression pins, not decoration.

| | before | after |
|---|---|---|
| PHPUnit | 851 | **879** |
| Jest | 2448 | **2455** |

PHPStan clean, PHPCS clean, build clean, no new `lint:js` errors vs `main`, pre-commit e2e (7 tests) pass.

## Not included

The audit's remaining Low findings are left for follow-up, none exploitable as written: merge-tag newlines in the email *subject* (PHPMailer's `secureHeader()` already strips them), `llms-txt` write paths lacking the `realpath()` containment the read paths have, and `map`'s `dsgoHeight` accepting a free-form CSS length.

Separately worth a product decision, not a security issue: the `/query/render` REST route requires `is_user_logged_in()`, so Dynamic Query load-more / infinite-scroll / AJAX filtering may not work for **logged-out** visitors. That's over-restrictive rather than under-, but it's probably not intended.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,8 @@ tasks.txt
 #Claude
 .claude/claude-memory.md
 .worktrees/
+
+# Security audit output. These enumerate unpatched findings with exploit paths,
+# so they must never land in a public repo — track the fixes, not the writeup.
+SECURITY-REVIEW.md
+SECURITY-AUDIT.md
--- a/includes/abilities/class-block-inserter.php
+++ b/includes/abilities/class-block-inserter.php
@@ -2046,8 +2046,20 @@ class Block_Inserter {
 	 * @return string Generated HTML.
 	 */
 	private static function generate_core_block_html( string $block_name, string $content, array $attributes ): string {
-		$text_align = isset( $attributes['textAlign'] ) ? ' has-text-align-' . $attributes['textAlign'] : '';
-		$align      = isset( $attributes['align'] ) ? ' has-text-align-' . $attributes['align'] : '';
+		// These land inside class="...", so an arbitrary value would break out of
+		// the attribute: textAlign of `center" onmouseover="alert(1)` yields a
+		// live event handler. They are alignment keywords, so rather than merely
+		// escape them, constrain them to the set core actually emits — then the
+		// value is provably safe rather than safe-looking. Anything else is
+		// dropped, which is also the correct rendering for a nonsense alignment.
+		$allowed_alignments = array( 'left', 'center', 'right', 'justify' );
+
+		$text_align = ( isset( $attributes['textAlign'] ) && in_array( $attributes['textAlign'], $allowed_alignments, true ) )
+			? ' has-text-align-' . $attributes['textAlign']
+			: '';
+		$align      = ( isset( $attributes['align'] ) && in_array( $attributes['align'], $allowed_alignments, true ) )
+			? ' has-text-align-' . $attributes['align']
+			: '';
 
 		switch ( $block_name ) {
 			case 'core/heading':
--- a/includes/abilities/class-css-sanitizer.php
+++ b/includes/abilities/class-css-sanitizer.php
@@ -23,6 +23,17 @@ if ( ! defined( 'ABSPATH' ) ) {
 class CSS_Sanitizer {
 
 	/**
+	 * Maximum encoding layers peeled before a value is rejected as unconverged.
+	 *
+	 * Real CSS carries zero or one layer of entity/escape encoding. Anything
+	 * needing more is layered deliberately, so the ceiling is generous for
+	 * legitimate input and still finite against an attacker choosing the depth.
+	 *
+	 * @var int
+	 */
+	private const MAX_DECODE_PASSES = 10;
+
+	/**
 	 * Dangerous CSS patterns that could be used for XSS attacks.
 	 *
 	 * @var array<string>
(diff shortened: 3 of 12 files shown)


jnealey-godaddy:
All three were right. Fixed in `4dc992bf` — and the High was a regression I introduced in the previous commit, so thank you for catching it.

### High — subject is a mail header, not HTML

Confirmed: `class-form-handler.php:880-881` fills **both** the subject and the body from the one map. The body is HTML and must be escaped; the subject is a plain-text header no mail client parses as markup, so escaping it isn't protection, it's corruption. And it's a documented merge-tag target, so this was a live authoring surface. My tests missed it because they only ever asserted on `$this->sent['message']`.

There are now two maps: `$body_tags` (esc_html'd) and `$subject_tags` (raw, newline-stripped so a submitted value can't smuggle in a second header — which also closes the audit's separate Low finding on subject newlines, rather than leaning on `PHPMailer::secureHeader()` downstream).

**Writing the test surfaced a second cause underneath the one you found.** With the maps split, the subject was *still* escaped — because **WordPress stores `blogname` already HTML-escaped** (`sanitize_option()` runs `esc_html` on save), so `get_bloginfo('name')` hands back `Bob&#039;s...` before we touch it. `{site_name}` now goes through `wp_specialchars_decode()`, which is exactly what core does when it puts the site name in an email. Splitting the maps alone would have left the reported symptom in place for the most common tag.

### Medium — decode ordering

Confirmed, and **a reorder alone isn't sufficient**. There are two independent encoding layers and each can hide the other, in *either* direction:

- `\3c script\3e` — a CSS escape hiding a tag from `wp_strip_all_tags()`
- `&#92;75rl(` — an entity hiding a CSS escape hiding `url(`

Whichever decode runs last exposes a layer the other already walked past, so no fixed sequence is safe. `sanitize()` now decodes entities + escapes to a **fixed point** (bounded loop), and only then strips tags and pattern-matches — on text with no encoding left to hide behind.

### Low

Extracted `Form_Handler::stringify_field_value()`.

### Verification

6 new tests, and I confirmed **each fails against the code as it stood before this commit**: 3 failures for the shared-map subject bug, and 2 for the CSS ordering bug (both escaped-tag payloads survived `sanitize()` — `A nested encoding survived sanitize(): \3c script\3e...`).

PHPUnit **885** (was 879), Jest 2455, PHPStan clean, PHPCS clean, build clean, e2e green.

jnealey-godaddy:
Both Lows fixed in `a501a375`. Both were in the shared boundary the rest of this PR now leans on, so neither gets to be fragile.

### `designsetgo_normalize_css_escapes()` — lone surrogates

Correct. The range check rejected `0` and `> 0x10FFFF` but not U+D800–U+DFFF. `mb_chr()` returns `false` for a surrogate, and `preg_replace_callback()` requires a string back — so `\d800` only works today because PHP coerces `false` to `''`. Now excluded explicitly, matching how the CSS spec treats them.

**Worth being straight about the test, though: it cannot fail against the old code.** The unguarded version produced the *same* output (`''`) via that coercion — same result, by accident. So there's no output-level assertion that distinguishes them, and I'm not going to claim a regression pin I don't have. Instead there's a second test pinning the platform fact underneath it:

```php
$this->assertFalse( mb_chr( 0xD800, 'UTF-8' ) );
```

That's the thing that would actually change if this ever bit us, so that's what's pinned. The guard makes correct behaviour *intentional* rather than incidental.

### `stringify_field_value()` — malformed shapes

Correct, and this one **does** have teeth. An array with no `value` key, or arrays nested inside arrays, fell through to `array_map( 'strval', … )`. It now recurses via `flatten_value()` and handles bool / null / non-scalar explicitly rather than assuming the happy shape — which is the whole point of the helper, given a submission is attacker-shaped data.

Verified: 2 `Array to string conversion` errors against the previous implementation, across 6 malformed-shape cases (no `value` key, nested arrays, deeply nested, bool, null, int).

### Verification

PHPUnit **896** (was 885), Jest 2455, PHPStan clean, PHPCS clean, build clean, e2e green.

jnealey-godaddy:
Both fixed in `5b3bc584`. The Medium was a real hole in the decode loop I added — thanks for running the nesting depth out.

### Medium — the cap returned a half-decoded string

Confirmed empirically that `html_entity_decode()` peels exactly one layer per call:

```
&amp;amp;amp;amp;amp;#106;avascript:alert(1)
 pass 5: &#106;avascript:alert(1)
 pass 6: javascript:alert(1)
```

My loop capped at 5 and then **returned what it had** — `&#106;avascript:alert(1)`. The pattern checks never see literal `javascript:`, so it survived into the "sanitized" output; inert in a `<style>` element, but a live vector the moment a caller drops it in an HTML attribute. And my comment claiming "the cap is a backstop, not a limit" was just wrong.

The fix isn't a bigger number — outrunnability is inherent to any bounded loop. It's the **behaviour at the ceiling**: non-convergence is now a **rejection** (`return '/* blocked */'`), not a partial best-effort. `MAX_DECODE_PASSES = 10` is generous for real CSS (zero or one layer); anything past it is deliberate.

This mirrors the defense-in-depth choice you noted `StyleBinding` already makes — reject on residual rather than trust the decode.

**Test has teeth**: without the reject branch, `sanitize()` returns `.a{color:red} &#106;avascript:alert(1)` — the exact residual — and the new deep-nesting test fails on it (6-layer and 12-layer cases).

### Low — docblock overclaim

Right, `normalize_css_escapes()`'s "not used for output" only held for `StyleBinding` (probe-only), not `CSS_Sanitizer` (the decoded form *is* its output). Reworded to describe both call patterns and state the actual contract: this function guarantees "no encoding left to hide behind", not "safe to echo" — a caller using the output runs its own pattern pass, as `CSS_Sanitizer` does.

### Verification

PHPUnit **898** (was 896), Jest 2455, PHPStan clean, PHPCS clean, build clean, e2e green.

jnealey-godaddy:
Caught up on all rounds. The latest full pass verified the decode-loop fix with no new findings — but going back over the earlier summaries, one Low was a real latent crash I hadn't addressed yet. Fixed in `b7383063`.

### Reply-To multi-value field → PHP 8 TypeError

You were right. When I added `stringify_field_value()` I applied it to the merge-tag path but missed the Reply-To extraction, which kept its own `is_array(...) ? ...['value'] : ...` — and that returns the inner value **unflattened**. If `emailReplyTo` points at a multi-value field, that inner value is an array, which then hits `sanitize_email()` → `strlen(array)` → fatal, killing the whole notification.

Routed through the same helper. The new test reproduces it exactly: against the old extraction it errors with

```
TypeError: strlen(): Argument #1 ($string) must be of type string, array given
```

### `%0a`/`%0d` in the subject strip

Added a comment explaining it, rather than removing it: it mirrors the header-injection strip at the top of the method (a value can arrive percent-encoded from an un-decoded source), and the trade-off — a subject legitimately containing the literal text "%0a" loses it — is negligible against a header strip.

### Left as-is

The `$fields` double-iteration (Low/DRY): the two loops build different structures — a keyed merge-tag map vs. two flat accumulators — so merging them costs more readability than the duplication does. Flagging that it's a deliberate keep, not an oversight.

### Verification

PHPUnit **899** (was 898), Jest 2455, PHPStan clean, PHPCS clean, build clean, e2e green.

PR #465, opened 2026-07-14
fix(pcp): add dsgo_ prefix to JS global and rate-limit transient key

ncimbaljevic-godaddy:
## Summary

- Rename `wp_localize_script` object `dsgoIntegrations` → `dsgo_integrations` in both PHP localization sites (`class-plugin.php`, `class-form-handler.php`) to satisfy the PCP `NonPrefixedVariableFound` sniff
- Update all JS consumers (`form-builder/view.js`, `MapSettingsPanel.js`) to match; bracket access replaced with `window.dsgo_integrations` + `eslint-disable-next-line camelcase` to satisfy both ESLint rules
- Prefix the rate-limit transient key with `dsgo_` in `class-form-security.php` (both `check_rate_limit` and `increment_rate_limit`) to avoid cross-plugin collisions
- Update the corresponding PHPUnit test fixture (`form-handler-test.php`) to match the new key

## Test plan

- [ ] Verify `npm run lint:js` passes (no camelcase / dot-notation errors)
- [ ] Verify `npm run lint:php` passes (no `NonPrefixedVariableFound` warnings)
- [ ] Verify Turnstile site key is still injected and read correctly in the form frontend
- [ ] Verify Google Maps API key still shows the "configured" notice in the Map block inspector
- [ ] Verify rate limiting still blocks after the configured max submissions (`test_rate_limit_exceeded` PHPUnit test)

### DIFF
--- a/includes/blocks/forms/class-form-handler.php
+++ b/includes/blocks/forms/class-form-handler.php
@@ -721,7 +721,7 @@ class Form_Handler {
 
 		wp_localize_script(
 			$handle,
-			'dsgoIntegrations',
+			'dsgo_integrations',
 			array(
 				'turnstileSiteKey' => ! empty( $integrations_settings['turnstile_site_key'] ) ? esc_js( $integrations_settings['turnstile_site_key'] ) : '',
 			)
--- a/includes/blocks/forms/class-form-handler.php
+++ b/includes/blocks/forms/class-form-handler.php
@@ -721,7 +721,7 @@ class Form_Handler {
 
 		wp_localize_script(
 			$handle,
-			'dsgo_integrations',
+			'dsgoIntegrations',
 			array(
 				'turnstileSiteKey' => ! empty( $integrations_settings['turnstile_site_key'] ) ? esc_js( $integrations_settings['turnstile_site_key'] ) : '',
 			)
--- a/includes/blocks/forms/class-form-security.php
+++ b/includes/blocks/forms/class-form-security.php
@@ -83,7 +83,7 @@ class Form_Security {
 	 */
 	public function check_rate_limit( string $form_id, int $block_max = 3 ) {
 		$ip_address = $this->get_client_ip();
-		$key        = 'dsgo_form_submit_' . md5( sprintf( '%s:%s', $form_id, $ip_address ) );
+		$key        = 'dsgo_form_submit_' . md5( $form_id ) . '_' . md5( $ip_address );
 		$count      = get_transient( $key );
 
 		$max_submissions = apply_filters( 'designsetgo_form_rate_limit_count', $block_max, $form_id );
@@ -109,7 +109,7 @@ class Form_Security {
 	 */
 	public function increment_rate_limit( string $form_id, int $block_window = 60 ): void {
 		$ip_address = $this->get_client_ip();
-		$key        = 'dsgo_form_submit_' . md5( sprintf( '%s:%s', $form_id, $ip_address ) );
+		$key        = 'dsgo_form_submit_' . md5( $form_id ) . '_' . md5( $ip_address );
 		$count      = get_transient( $key );
 
 		$time_window = apply_filters( 'designsetgo_form_rate_limit_window', $block_window, $form_id );
--- a/includes/blocks/forms/class-form-security.php
+++ b/includes/blocks/forms/class-form-security.php
@@ -83,7 +83,7 @@ class Form_Security {
 	 */
 	public function check_rate_limit( string $form_id, int $block_max = 3 ) {
 		$ip_address = $this->get_client_ip();
-		$key        = 'dsgo_form_submit_' . md5( $form_id ) . '_' . md5( $ip_address );
+		$key        = 'dsgo_form_submit_' . md5( sprintf( '%s:%s', $form_id, $ip_address ) );
 		$count      = get_transient( $key );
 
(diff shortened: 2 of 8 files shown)


jnealey-godaddy:
## Review: split this — the transient half is good, the JS-global half doesn't do what it needs to

### ✅ Rate-limit transient prefix — merge as-is

`dsgo_form_submit_` + `md5(form_id)` + `md5(ip)` is right, and it lines up with the plugin's existing `dsgo_has_blocks_` / `dsgo_form_attrs_v2_` / `dsgo_form_submissions_count` keys. `form_submit_*` was a genuinely collision-prone option-row name.

Verified end-to-end against the real REST endpoint (`designsetgo/v1/form/submit`) in wp-env:

```
#1: HTTP 200 :: Form submitted successfully!
#2: HTTP 200 :: Form submitted successfully!
#3: HTTP 200 :: Form submitted successfully!
#4: HTTP 429 :: Too many submissions. Please try again later.
#5: HTTP 429 :: Too many submissions. Please try again later.
```

DB afterwards holds exactly one correctly-prefixed row (`_transient_dsgo_form_submit_<md5>_<md5>`). All five key-construction sites are updated consistently (both methods in `class-form-security.php`, the PHPUnit fixture, both patterns in `uninstall.php`) — no missed call site. `phpunit --filter "rate_limit|uninstall"` passes (8 tests, 22 assertions). Hashing `form_id` is a nice bonus: it bounds the key length, which matters because `option_name` is `varchar(191)` and `form_id` is author-controlled.

**Nit:** `uninstall.php` now only matches `_transient_dsgo_form_submit_%`, so sites upgrading from a previous version keep their legacy `_transient_form_submit_%` rows (I saw 3 left behind in my DB). Core's daily `delete_expired_transients` cron reaps them, so it self-heals — but keeping the old LIKE pattern for a release costs one line.

### ❌ `dsgoIntegrations` → `dsgo_integrations` — doesn't satisfy the wp.org feedback

The underlying concern is legitimate: `wp_localize_script` really does create a `window`-level global, so it's in scope for wp.org's "prefix everything in the global namespace" rule. But this specific change doesn't answer it, for two reasons.

**1. Snake_case is orthogonal to prefixing.** `dsgoIntegrations` and `dsgo_integrations` carry the *same* `dsgo` prefix — the rename changes the casing, not the prefix. If the objection was "this global isn't prefixed," it isn't any more prefixed now. The wp.org rule is about prefix uniqueness, not word separators; core itself ships camelCase localized globals (`wpApiSettings`, `wpCookies`).

For what it's worth, the PR description attributes this to the PCP `NonPrefixedVariableFound` sniff, and that attribution is wrong — that sniff (WPCS `PrefixAllGlobalsSniff`) only ever fires on PHP **global variables** (its message template is literally `'Global variables defined'`); `wp_localize_script` isn't in its `target_functions`. A probe file containing a completely unprefixed `wp_localize_script( 'h', 'integrations', [...] )` passes it clean, and running it across the whole plugin reports zero findings. Its matcher is also `stripos( $name, $prefix ) === 0` — case-insensitive, no separator requirement — so even hypothetically, `dsgoIntegrations` already satisfies a `dsgo` prefix. So this needs to be justified by the human review feedback, not the sniff.

**2. It fixes one global out of eleven.** The plugin puts all of these on `window`:

| Global | Location |
|---|---|
| `dsgoIntegrations` | `class-plugin.php:834`, `class-form-handler.php:724` ← the only one this PR touches |
| `dsgoSettings` | `class-plugin.php:884` |
| `dsgoStickyHeaderGlobalSettings` | `class-plugin.php:821` |
| `dsgStickyHeaderSettings` | `class-sticky-header.php:99` (only a 3-char `dsg` prefix) |
| `dsgoIcons` / `dsgoIconAliases` / `dsgoIconDefaults` | `class-icon-injector.php:101/108/117` |
| `dsgoQueryFilterIndexDashboard` | `class-query-filter-index-admin.php:101` |
| `dsgoDeactivation` | `class-block-migrator.php:204` |
| `designsetgoForm` | `class-form-handler.php:708` |
| `designSetGoAdmin` | `class-admin-menu.php:223` |

If `dsgo` is an acceptable prefix to wp.org, `dsgoIntegrations` was already fine and there's nothing to do. If it isn't, the other ten are equally exposed. Either way, renaming one of them to snake_case doesn't get the plugin to pass — and it breaks any theme/snippet reading `window.dsgoIntegrations` while contradicting our own convention (`dsgoAttributeName` for JS, per CLAUDE.md).

**Suggested fix:** standardize all eleven on the full slug prefix in camelCase (`designsetgoIntegrations`, `designsetgoSettings`, …), or consolidate them under one `designsetgo` namespace object — as a separate, complete pass. Worth pasting the exact wp.org email into that PR, since the symbols they named determine whether `dsgo` is acceptable at all.

Worth keeping from this PR regardless: the optional-chaining cleanup in `view.js` (`window.dsgoIntegrations?.turnstileSiteKey ?? null`) is a genuine improvement — just keep the camelCase name.

### QA notes

Smoke-tested in wp-env (WP 6.9 / PHP 8.3) against the PR branch with a rebuilt `build/`:

- Editor: `window.dsgo_integrations` populates; Map inspector renders both the "No API key configured" warning and the "✓ API key configured" success branch correctly.
- Frontend: with Cloudflare's official test keys, Turnstile renders, hits "Success!", and populates `dsg_turnstile_token`.
- No JS console errors attributable to the rename.

jnealey-godaddy (approved):

PR #467, opened 2026-07-14
fix(security): port stranded 2.2.0 hardening fixes to main

jnealey-godaddy:
Two security fixes were merged into `release/2.2.0` but never reached `main`. PR #433 was opened with `release/2.2.0` as its base rather than `main`, so the work stayed on the release branch. The `v2.2.0` tag is an ancestor of `main`, which made the branch look fully retired — but its tip carries six commits beyond the tag, two of them security hardening.

Confirmed via `git log -S` that neither `MAX_OBJECT_SCAN_DEPTH` nor `LEGACY_REDACTED_PLACEHOLDER` has ever existed anywhere in `main`'s history, so this is a genuine gap rather than a deliberate revert.

## What this ports

**`contains_object()` recursion cap** (`includes/admin/class-draft-mode.php`)
`main`'s version recurses unbounded while scanning unserialized meta for object-injection payloads. A deeply nested array can exhaust the stack. This adds `MAX_OBJECT_SCAN_DEPTH = 20` and fails closed — anything deeper is treated as containing an object, so the caller skips it.

**Secret-overwrite window** (`includes/admin/class-settings.php`)
The redaction placeholder becomes ASCII `__DSGO_REDACTED__`, byte-identical across encodings and opcode caches. A browser holding a settings form cached from before the upgrade echoes the old `••••••••` string back on submit; the write path now accepts `LEGACY_REDACTED_PLACEHOLDER` as "unchanged" via `is_redaction_placeholder()`, so that stale submit preserves the stored secret instead of overwriting it with literal bullets. Read paths only ever emit the current placeholder.

## Compatibility

Changing `REDACTED_PLACEHOLDER`'s value is safe: nothing in `src/` compares against the literal, and the only remaining reference to the bullet string is the migration note in the docblock. The admin UI round-trips whatever the REST GET returns.

## Verification

- `npm run lint:php` — clean, 117/117 files
- Full PHPUnit suite — **909 tests, 4206 assertions, all passing**
- Mutation-checked the depth cap: removing the guard makes `test_create_draft_skips_meta_nested_beyond_depth_cap` fail, and restoring it passes, so the test genuinely covers the fix

Both cherry-picks retain original authorship (Nemanja Cimbaljevic). The only conflict was a docblock in `class-settings.php`; the code merged cleanly.

Once this lands, `release/2.2.0` holds nothing unique and can be deleted.

### DIFF
--- a/includes/admin/class-draft-mode.php
+++ b/includes/admin/class-draft-mode.php
@@ -36,6 +36,14 @@ class Draft_Mode {
 	const META_DRAFT_CREATED = '_dsgo_draft_created';
 
 	/**
+	 * Maximum nesting depth scanned by contains_object().
+	 *
+	 * Anything deeper is treated as containing an object (fail closed),
+	 * capping recursion so a maliciously deep payload cannot overflow the stack.
+	 */
+	const MAX_OBJECT_SCAN_DEPTH = 20;
+
+	/**
 	 * Constructor
 	 */
 	public function __construct() {
@@ -472,17 +480,27 @@ class Draft_Mode {
 	 * serialized payload like `a:1:{i:0;O:8:"stdClass":0:{}}`), which would
 	 * still trigger object injection when stored and later unserialized.
 	 *
+	 * Recursion is bounded by self::MAX_OBJECT_SCAN_DEPTH. A structure deeper
+	 * than that is treated as if it contains an object (fail closed) so the
+	 * caller skips it — both eliminating any stack-overflow risk from a
+	 * maliciously deep payload and erring on the side of rejecting it.
+	 *
 	 * @param mixed $value Value to inspect (already passed through maybe_unserialize()).
+	 * @param int   $depth Current recursion depth (internal).
 	 * @return bool True if the value is an object or any nested array element is.
 	 */
-	private static function contains_object( $value ) {
+	private static function contains_object( $value, $depth = 0 ) {
 		if ( is_object( $value ) ) {
 			return true;
 		}
 
 		if ( is_array( $value ) ) {
+			if ( $depth >= self::MAX_OBJECT_SCAN_DEPTH ) {
+				return true;
+			}
+
 			foreach ( $value as $item ) {
-				if ( self::contains_object( $item ) ) {
+				if ( self::contains_object( $item, $depth + 1 ) ) {
 					return true;
 				}
 			}
--- a/includes/admin/class-draft-mode.php
+++ b/includes/admin/class-draft-mode.php
@@ -36,12 +36,29 @@ class Draft_Mode {
 	const META_DRAFT_CREATED = '_dsgo_draft_created';
 
 	/**
-	 * Maximum nesting depth scanned by contains_object().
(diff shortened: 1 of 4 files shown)


jnealey-godaddy:
Thanks — chasing this down turned up something worse than the silent skip, so the comment was load-bearing.

## The skip wasn't just silent, it was destructive

`MAX_OBJECT_SCAN_DEPTH` fails closed, so a legitimately deep but object-free value (exactly the nested ACF/page-builder case you raised) never reaches the draft. `sync_post_meta()` then read that absence from the draft as *"the author deleted this key"* and synced the deletion back:

```php
if ( ! isset( $draft_meta[ $key ] ) ) {
    delete_post_meta( $original_id, $key );   // key was never copied, not deleted by the author
}
```

So publishing a draft **permanently deleted deep meta from the original** — meta that predated the draft and that the author never touched. I confirmed it with a probe before fixing: a 40-level object-free value was withheld from the draft and then wiped from the original on publish, while a shallow key alongside it survived. Since the depth cap is new in this PR, the port introduced that regression — `main`'s unbounded scanner copied such values fine.

## Fix

The two rejection reasons mean different things, so they're now distinguished (`REJECT_OBJECT` / `REJECT_DEPTH`). An object is hostile and is never copied anywhere. "Too deep to vet" says nothing about whether the value is safe, so that key is now **left untouched rather than deleted**. I also moved vetting ahead of `delete_post_meta()` so a key with even one unwritable value isn't deleted and then left unwritten.

## Your actual ask: debuggability

Skips now fire `designsetgo_draft_meta_skipped` (key, reason, post ID, outcome) and log under `WP_DEBUG`, naming the key and reason but never the value, which may hold a secret. Real log output from the smoke test:

```
DesignSetGo: meta key "acf_deep_config" on post 2706 was not copied because it nests
deeper than the 20-level scan cap; the draft will not carry this key.
DesignSetGo: meta key "acf_deep_config" on post 2706 was not copied because it nests
deeper than the 20-level scan cap; it was left untouched on the original rather than deleted.
```

I went with `error_log()` over `wp_trigger_error()`: the latter raises an `E_USER_NOTICE`, which PHPUnit converts to an exception and would fail every test that legitimately skips a key. The gated `error_log()` form already matches `class-form-handler.php`.

## Verification

- Full PHPUnit suite: **911 tests, 4213 assertions, green** (up from 909 — two new regression tests)
- `npm run lint:php`: clean, 117/117
- Both new tests mutation-checked — neutering the preserve guard makes `test_publish_draft_preserves_meta_too_deep_to_copy` fail, so it genuinely catches the data loss
- Browser smoke test against wp-env (Chrome DevTools): created a page with 30-level meta, ran Create Draft → edit → Publish Changes through the real UI. Content synced, the deep meta survived with all 30 levels intact, draft cleaned up, no console errors.
- Also exercised the settings fix end-to-end: stored a real secret, confirmed the REST GET never leaks it, then wrote back **both** the current and the legacy `••••••••` placeholder — the stored secret survived both, verified directly in the database.

PR #468, opened 2026-07-15
fix(draft-mode): preserve excluded meta on publish; make depth cap filterable

jnealey-godaddy:
Follow-up to #467, addressing the two review comments left on it. Both were flagged Low/non-blocking, but the first is confirmed data loss.

## 1. Excluded meta was destroyed on publish (data loss)

`copy_post_meta()` skips keys in `designsetgo_draft_excluded_meta_keys`, but `sync_post_meta()` never consulted that filter. So a key an integrator deliberately kept off the draft was present on the original yet absent from the draft at publish time — which the publish step read as *"the author deleted it"* and synced the deletion:

```php
if ( ! isset( $draft_meta[ $key ] ) ) {
    delete_post_meta( $original_id, $key );   // excluded, never copied — not deleted by the author
}
```

A third-party integration keeping its own bookkeeping meta off the draft therefore lost that key on **every** publish. This is the same shape as the depth-cap bug fixed in #467, reached by a different route: the value is perfectly copyable, so the object/depth guard doesn't catch it.

I confirmed it with a probe (an excluded key, seeded on a page, gone from the original after a create-draft → publish cycle) before fixing.

**Fix:** both functions now read the excluded-keys list from one shared `get_excluded_meta_keys()` helper, so the publish step recognises an excluded key's absence from the draft as "never copied" rather than "removed."

## 2. Depth cap is now filterable (reviewer request)

`MAX_OBJECT_SCAN_DEPTH` was a hardcoded 20 with no escape hatch, so a site with legitimately deep but object-free data (deep ACF repeater / flexible-content or page-builder configs) had it withheld from every draft with no recourse.

Added `designsetgo_draft_max_object_scan_depth`, mirroring the existing `designsetgo_draft_excluded_meta_keys` filter. The result is **clamped to `ABSOLUTE_MAX_OBJECT_SCAN_DEPTH` (200)** — the cap exists to bound recursion against a hostile payload, so the filter may relax it but must not be able to hand that guarantee away. The filter runs once per scan (resolved on entry, threaded down the recursion), not once per nested element.

## Not changed

The pre-existing split between `designsetgo_draft_excluded_meta_keys` (copy) and `designsetgo_draft_preserve_meta_keys` (publish) is left as-is — unifying them would change public filter semantics. This PR only stops the copy-side exclusion from causing deletion, which is the actual bug.

## Verification

- Full PHPUnit suite: **914 tests, 4222 assertions, green** (three new tests)
- `npm run lint:php`: clean, 117/117
- All three new tests mutation-checked:
  - `test_publish_draft_preserves_excluded_meta_keys` — fails if the shared-exclusion guard is removed
  - `test_scan_depth_filter_allows_deeper_meta` — 25-level value copies when the filter raises the cap to 40
  - `test_scan_depth_filter_is_clamped_to_absolute_max` — a value past the ceiling stays rejected even when the filter returns `PHP_INT_MAX`

### DIFF
--- a/includes/admin/class-draft-mode.php
+++ b/includes/admin/class-draft-mode.php
@@ -36,12 +36,23 @@ class Draft_Mode {
 	const META_DRAFT_CREATED = '_dsgo_draft_created';
 
 	/**
-	 * Maximum nesting depth scanned by meta_rejection_reason().
+	 * Default maximum nesting depth scanned by meta_rejection_reason().
 	 *
 	 * Anything deeper is rejected (fail closed), capping recursion so a
-	 * maliciously deep payload cannot overflow the stack.
+	 * maliciously deep payload cannot overflow the stack. Raise it for a
+	 * specific site with the designsetgo_draft_max_object_scan_depth filter.
 	 */
 	const MAX_OBJECT_SCAN_DEPTH = 20;
+
+	/**
+	 * Hard ceiling on the scan depth, whatever the filter asks for.
+	 *
+	 * The depth cap's whole purpose is to bound recursion, so the filter that
+	 * relaxes it must not be able to remove that bound entirely. This sits far
+	 * above any plausible real structure and far below a stack overflow, so the
+	 * guarantee holds no matter what an integrator passes.
+	 */
+	const ABSOLUTE_MAX_OBJECT_SCAN_DEPTH = 200;
 
 	/**
 	 * Rejection reason: the value is, or contains, a PHP object.
@@ -365,23 +376,7 @@ class Draft_Mode {
 	private function copy_post_meta( $source_id, $target_id ) {
 		$meta = get_post_meta( $source_id );
 
-		$excluded_keys = array(
-			'_edit_lock',
-			'_edit_last',
-			self::META_DRAFT_OF,
-			self::META_HAS_DRAFT,
-			self::META_DRAFT_CREATED,
-			'_wp_old_slug',
-			'_wp_old_date',
-		);
-
-		/**
-		 * Filter the meta keys to copy when creating a draft.
-		 *
-		 * @param array $excluded_keys Keys to exclude from copying.
-		 * @param int   $source_id     Source post ID.
-		 */
-		$excluded_keys = apply_filters( 'designsetgo_draft_excluded_meta_keys', $excluded_keys, $source_id );
+		$excluded_keys = self::get_excluded_meta_keys( $source_id );
 
 		foreach ( $meta as $key => $values ) {
 			if ( in_array( $key, $excluded_keys, true ) ) {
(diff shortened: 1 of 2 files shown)


jnealey-godaddy:
Good catch, and it's real — the deletion decision re-evaluates the filter at publish time, so a callback that isn't a pure function of `$post_id` can exclude a key at copy time and admit it at publish time, at which point the deletion sync destroys it.

Worth adding: the **same instability applies to `designsetgo_draft_max_object_scan_depth`**, the filter this PR introduces — and there the trigger is more mundane than an oddly-designed callback. A deploy that raises the cap while a draft is open makes a previously-too-deep key vet as copyable on publish, and it gets deleted.

I went with the docblock contract you suggested, on **both** filters (013615a3): they must be stable across a draft's create→publish lifetime, with the reason spelled out so the footgun is documented at the point of extension. Docs-only, no behaviour change, tests still green (68 Draft_Mode / 914 total), lint clean.

I considered the fully-robust alternative — snapshotting the withheld keys onto the draft at creation and reading that back on publish instead of re-evaluating the filters, which would close the class outright (both filters + any future rejection reason). Held off: it adds persistent bookkeeping meta plus a migration fallback for in-flight drafts, which is more than this approved PR should carry for an edge case that needs a deliberately unstable filter. Happy to do it as a follow-up if you'd rather eliminate the contract dependency entirely.

jnealey-godaddy:
Fixed in cf4cb036 — you were right, and I'd measured it wrong the first time by not mirroring the `$excluded_keys` pattern for the depth filter.

`copy_post_meta()` and `sync_post_meta()` now resolve `get_max_object_scan_depth()` once at the top (right next to `$excluded_keys`) and thread the value through `meta_rejection_reason()`, `values_rejection_reason()` and `log_skipped_meta()`. Each helper keeps a null-sentinel fallback so it stays correct if called standalone.

Quantified it with a probe — a page with 24 meta values:

| | depth-filter dispatches on create_draft |
|---|---|
| before | 25 |
| after | 1 |

Publish is likewise one dispatch. No behaviour change; full suite still green (914 tests, 4222 assertions) and lint clean.

jnealey-godaddy:
Fixed in 7a43ada5. Reworded all three spots to describe the bound as the resolved cap rather than the constant:

- the `REJECT_DEPTH` constant docblock
- the `meta_rejection_reason()` docblock ("Recursion is bounded by…")
- the `designsetgo_draft_meta_skipped` action docblock ("'depth' (nests deeper than…)")

Each now points at `get_max_object_scan_depth()` and notes that `MAX_OBJECT_SCAN_DEPTH` is only the default. (The constant's own docblock already said "Default … Raise it with the filter", so it was left as-is.) Docs-only, lint clean.

PR #469, opened 2026-07-15
feat(form-builder): submit button style variations (secondary/outline)

jnealey-godaddy:
## Summary

Adds a **Button Style** control to the Form Builder submit button so a form placed on an alternate background can use a secondary/outline button, and so section/group style variations can restyle it via a class hook.

- New `submitButtonVariation` attribute (`default` | `secondary` | `outline`) emits `dsgo-form__submit--{variation}` on the submit `<button>` (both inline and below positions).
- Inspector **Button Style** select in the Settings panel, with editor-preview parity and reset support.
- Self-contained "ghost" fallback in `style.scss` (transparent + current-color border, inherited text) so secondary/outline read on any background without a Style Kit. Kits — and section/group style variations — can repaint the same classes.

## Why this is the fix

The submit button is a hardcoded `<button>` baked into `save()` with no `className` passthrough, so a block-style variation (`is-style-*`) can't attach to it and gets stripped on invalid-block recovery. A first-class attribute is the correct per-instance mechanism; the wrapper-cascade path (section/group variations) remains available too.

## Specificity note

The `designsetgo` theme styles `.dsgo-form__submit` as a filled primary button at `(0,3,0)` via `class-button-global-styles.php`. The fallback is compounded to `(0,4,0)` (`:root` + three element classes) to win — mirroring the icon-button `is-style-outline` approach.

## No deprecation required

Verified in a running editor: the default serialized output is **byte-identical** to the prior output (`class="dsgo-form__submit wp-element-button"`, and `submitButtonVariation` is omitted from the block comment at its default), so existing forms stay valid. Adding a deprecation would be dead code.

## Verification

- Default / secondary / outline all round-trip `isValid: true`; full `deprecations-isEligible` suite passes (218/218).
- Rendered on the real frontend: the secondary button computes `background: transparent`, `border: 1px solid currentColor`, inherited text (ghost button) instead of the filled default.
- `lint:js` and `lint:css` clean; `npm run build` succeeds.

## Test plan

- Insert a Form Builder block → Settings → **Button Style** → Secondary/Outline; confirm editor preview and frontend both show the ghost button.
- Confirm an existing form (no Button Style set) is unchanged and does not prompt "Attempt Recovery".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - **Cloudflare Turnstile no longer rejects every submission it protects.** Turnstile was unusable on any site that enabled it: the form endpoint validated `turnstile_token` against `/^[a-zA-Z0-9_-]+$/`, which has no `.` in the character class, but real Turnstile tokens are dot-delimited. The widget rendered and issued its token normally, then every submission failed REST parameter validation with `400 Invalid parameter(s): turnstile_token` before reaching the handler — so enabling Turnstile silently broke the form instead of protecting it. The token is opaque (Cloudflare guarantees only a 2048-character ceiling and does not contract its alphabet), so the validator no longer tries to parse its structure: it bounds the length and rejects only what a token can never contain — whitespace, newlines, and other non-printable bytes — before forwarding the value to `siteverify`. Non-string input is now rejected cleanly rather than reaching `preg_match()` as a PHP 8 `TypeError`. Pre-existing in the shipped 2.4.0 tag, and backed by a regression test that dispatches the real route and fails against the unpatched code. (#466)
 
+### New Features
+- **Form Builder: submit button style (Secondary / Outline)** — The submit button gains a Button Style control (`default` / `secondary` / `outline`) that emits a `dsgo-form__submit--{variation}` class, so a form placed on an alternate background can use a secondary/outline button and section/group style variations can restyle it. Ships a self-contained ghost fallback (transparent + current-color border) scoped above the theme's filled-button rule; Style Kits repaint the same class. Default output is byte-identical, so existing forms are untouched and need no migration. (#469)
+
 ## [2.4.0] - 2026-07-12
 
 ### Security
--- a/includes/abilities/class-block-inserter.php
+++ b/includes/abilities/class-block-inserter.php
@@ -1899,6 +1899,7 @@ class Block_Inserter {
 		$submit_button_text               = $attributes['submitButtonText'] ?? 'Submit';
 		$submit_button_alignment          = $attributes['submitButtonAlignment'] ?? 'left';
 		$submit_button_position           = $attributes['submitButtonPosition'] ?? 'below';
+		$submit_button_variation          = $attributes['submitButtonVariation'] ?? 'default';
 		$ajax_submit                      = $attributes['ajaxSubmit'] ?? true;
 		$success_message                  = $attributes['successMessage'] ?? 'Thank you! Your form has been submitted successfully.';
 		$error_message                    = $attributes['errorMessage'] ?? 'There was an error submitting the form. Please try again.';
@@ -1923,6 +1924,12 @@ class Block_Inserter {
 		$email_from_email                 = $attributes['emailFromEmail'] ?? '';
 		$email_reply_to                   = $attributes['emailReplyTo'] ?? '';
 		$email_body                       = $attributes['emailBody'] ?? '';
+
+		// Submit-button style variation class - must match save.js. Validated
+		// against the block.json enum so an AI-supplied value can't inject markup.
+		$submit_button_variation_class = in_array( $submit_button_variation, array( 'secondary', 'outline' ), true )
+			? ' dsgo-form__submit--' . $submit_button_variation
+			: '';
 
 		// Build classes - must match save.js.
 		$classes = $block_class;
@@ -1999,7 +2006,7 @@ class Block_Inserter {
 
 		// Inline button goes inside fields wrapper, before closing.
 		if ( 'inline' === $submit_button_position ) {
-			$closing .= '<button type="submit" class="dsgo-form__submit dsgo-form__submit--inline wp-element-button" style="' . $button_style . '">' . esc_html( $submit_button_text ) . '</button>';
+			$closing .= '<button type="submit" class="dsgo-form__submit dsgo-form__submit--inline' . $submit_button_variation_class . ' wp-element-button" style="' . $button_style . '">' . esc_html( $submit_button_text ) . '</button>';
 		}
 
 		// Close fields wrapper.
@@ -2021,7 +2028,7 @@ class Block_Inserter {
 		// Footer with button (below position).
 		if ( 'below' === $submit_button_position ) {
 			$closing .= '<div class="dsgo-form__footer">';
-			$closing .= '<button type="submit" class="dsgo-form__submit wp-element-button" style="' . $button_style . '">' . esc_html( $submit_button_text ) . '</button>';
+			$closing .= '<button type="submit" class="dsgo-form__submit' . $submit_button_variation_class . ' wp-element-button" style="' . $button_style . '">' . esc_html( $submit_button_text ) . '</button>';
 			$closing .= '</div>';
 		}
 
--- a/src/blocks/form-builder/block.json
+++ b/src/blocks/form-builder/block.json
@@ -45,6 +45,15 @@
 				"inline"
 			]
 		},
+		"submitButtonVariation": {
+			"type": "string",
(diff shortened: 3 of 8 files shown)


jnealey-godaddy:
Thanks — both addressed in `bb2109e`:

**High (hover-bg guard collision):** confirmed real and fixed. The `:not([class*="dsgo-form__submit--"])` guard was a substring proxy for "no animation class," and the new `--secondary`/`--outline` variations share that prefix. Rather than rename the variation classes (which would break BEM and ripple into the generator prototype #470), I replaced the substring with the explicit animation set — `:not(.dsgo-form__submit--fill-diagonal, --zoom-in, --lift, --shrink, --border-pulse, --border-glow, --slide-{left,right,down,up})` — kept in sync with `ALLOWED_HOVER_ANIMATIONS`. Verified via `element.matches()`: a Secondary button with `--dsgo-button-hover-bg` and no animation was excluded by the old guard (`false`) and now matches (`true`), while a `--zoom-in` button stays correctly excluded (handled by its own rule). This also removes the latent landmine for any future non-animation `dsgo-form__submit--*` modifier.

**Low (CHANGELOG):** added an `[Unreleased] → New Features` entry referencing #469.

jnealey-godaddy:
Thanks for the follow-up review.

**Low (`--loading`):** good catch — fixed in `66bc153`. Added `.dsgo-form__submit--loading` to the exclusion list so the spinner/disabled state never flashes the custom hover background, exactly as the old substring guard did incidentally.

**Medium (DRY — third copy of the animation slugs):** acknowledged, and I'll keep it as-is for this PR. It's consistent with the existing convention rather than a new violation — `edit.js`'s `ALLOWED_ANIMATIONS` is already a second hardcoded copy of `class-plugin.php`'s `ALLOWED_HOVER_ANIMATIONS` for the same editor/frontend-parity reason. The list only changes when an animation is added, which already requires touching this file (each animation has its own transition rule here), and the comment flags the sync requirement. If we want real enforcement, the right move is a single follow-up that makes all three consumers read one source of truth (e.g. generate the CSS list) rather than bolting a cross-language assertion onto this feature PR — happy to open that separately if you'd like.

jnealey-godaddy:
Good catch — fixed in \`b56080d\`. Added \`.dsgo-form__submit--no-hover\` to the exclusion list too. Confirmed it's a real class (styled in \`style.scss:558\`, and \`apply_default_form_button_hover()\` at \`class-plugin.php:1132\` treats it as an explicit opt-out), so a no-hover button carrying a custom hover-bg must not flash the colour. The exclusion list is now a faithful superset of everything the old substring guard incidentally covered (animations + \`--loading\` + \`--no-hover\`).

jnealey-godaddy:
Confirmed and fixed in \`b9ad1c1\` — and verifying it turned up that the problem is **broader** than the boosted-specificity rule.

**What I found:** the loading state hides the label with \`color: transparent\` at \`.dsgo-form__submit.dsgo-form__submit--loading\` = **(0,2,0)**. That already loses to \`Button_Global_Styles\`' base rule at **(0,3,0)** — so on any kit that sets a button text colour, the default submit button *already* keeps its text visible during loading (reproduced: \`loadingColor\` stayed white, text not hidden). My variation fallback at (0,4,0) made it worse for Secondary/Outline, but the base case is broken on \`main\` too. So the suggested \`:not(.dsgo-form__submit--loading)\` on the fallback would fix Secondary/Outline but leave the default button broken.

**Why not just \`!important\` the colour:** the spinner (\`::after\`) paints with \`currentColor\` via \`color: inherit\`, so forcing the button \`color: transparent\` hides the text **and blanks the spinner**. (That's a pre-existing latent bug: on a plain button where \`color: transparent\` *did* apply, the spinner was already invisible.)

**The fix:** hide the label with \`text-indent\` instead of \`color\`. Nothing else sets \`text-indent\`, so there's no specificity war, and \`currentColor\` stays real so the spinner is visible. This fixes the default button, Secondary, and Outline uniformly — and the pre-existing spinner-blanking case.

Verified the mechanics in an isolated repro (label pushed off-screen, \`::after\` border keeps a real colour). Heads-up: a couple of local wp-env instances collided on the port mid-review, so I couldn't drive a fresh live AJAX submit — the isolated repro replicates the exact rule interaction (block button + generated 0,3,0/0,4,0 colour + loading class + spinner). One cosmetic note: a shrink-to-fit submit button narrows slightly while loading (the label no longer contributes width); acceptable for a sub-second state, and better than text overlapping the spinner.

jnealey-godaddy:
Addressed both non-blocking observations:

**Block inserter parity** (\`4c9e7b2\`): \`generate_form_builder_html()\` now reads \`submitButtonVariation\` and emits \`dsgo-form__submit--{variation}\`, validated against the block.json enum (\`secondary\`/\`outline\`; anything else — including a class-attribute breakout attempt — is dropped). Class order matches save.js, so an AI-inserted Secondary/Outline form is valid content and a default form stays byte-identical.

**Test coverage** (\`a6fb5d7\`): added \`Block_Inserter_Form_Variation_Test\` (7 tests) covering default/secondary/outline, inline ordering, and the enum-rejection/breakout cases. That's the PHP generation path; on the JS save side, \`tests/unit/deprecations-isEligible.test.js\` already round-trips form-builder across all attribute variants (now including \`submitButtonVariation\`), so the attribute's serialize→parse validity is covered even though the block has no bespoke JS unit tests.

jnealey-godaddy:
Latest round addressed:

**Medium — ghost fallback fills on hover** (\`1b627b4\`): fixed. Added a \`:hover\` rule at 0,5,0 for \`--secondary\`/\`--outline\` so the ghost stays transparent, beating the base Global-Styles hover rule (0,4,0). Now deterministic (was source-order dependent). A kit that *wants* a hover fill for these defines it under \`core/button.variations.{name}:hover\`, which the generator (#470) projects at matching specificity.

**Low — redundant \`overflow: hidden\`** (\`1b627b4\`): removed; it's already on the parent \`.dsgo-form__submit\` rule.

**Low — inline \`submitButtonBackgroundColor\` beats the ghost \`transparent\`:** correct, and left as-is. Inline styles always beat class selectors, so a button with an explicit custom background that's *also* switched to Secondary/Outline keeps the custom background. That's a contradictory author choice (a hand-picked fill vs. a ghost variation); rather than silently drop the author's custom colour, the custom colour wins. Worth a UI hint later, but not something CSS should paper over.

**Low — explicit exclusion list un-excludes \`--inline\`:** intentional. \`--inline\` is a layout modifier, not a hover animation — an inline-position submit button with a custom hover-bg and no animation *should* get the instant hover background. The old substring guard excluded it incidentally (a latent miss); the explicit list correctly excludes only the animation/loading/no-hover states, so this is a fix, not a regression.

jnealey-godaddy:
Both addressed in \`5a91cc9\`.

**Medium — custom color silently beats the ghost look:** added an explicit note to the Button Style control's help text — *"A custom Button Background/Text Color (in the Color panel) is applied inline and takes precedence over Secondary/Outline — clear it to use the ghost style."* This removes the "silently solid, no explanation" surprise. I kept the precedence as-is (custom colour wins) rather than silently dropping a hand-picked colour when a variation is selected; the help text makes the interaction explicit, and full mutual-exclusion UI (disabling the colour controls when a variation is active) is a reasonable follow-up if we want it.

**Low — CSS `:not()` list duplicates `ALLOWED_HOVER_ANIMATIONS` with no enforcement:** added \`Form_Builder_Hover_Guard_Sync_Test\`. It reads the \`:hover:not( … )\` guard out of \`style.scss\` and asserts every \`Plugin::ALLOWED_HOVER_ANIMATIONS\` slug is present, so adding an animation to the PHP constant without updating the CSS guard now fails CI (with a message naming the missing slug). That's the enforcement that was missing across the three copies.

jnealey-godaddy:
Thanks for the approval. Both notes addressed in \`e472329b\`.

**Medium — ghost \`:hover\` comment overclaims "deterministic":** correct, the comment was wrong. The instant-hover-bg rule below (\`--dsgo-button-hover-bg !important\`) does still win over the 0,5,0 ghost \`:hover\`, so a per-instance custom hover colour beats the ghost. Fixed the comment to say so explicitly — and that's the *intended* precedence: it mirrors the base custom-colour behavior (a hand-picked colour beats the variation, per the Button Style help text). The 0,5,0 rule only guarantees the ghost stays transparent against the *base Global-Styles* hover fill (0,4,0), which is the actual case it's there for.

**Low — no JS enum validation:** added. \`save.js\` and \`edit.js\` now allowlist \`submitButtonVariation\` against the block.json enum (\`['secondary','outline']\`) before interpolating it into the class, matching \`class-block-inserter.php\`. React already escapes the className so there was no breakout risk, but this makes the three code paths consistent and keeps a stray attribute value out of the class entirely. Byte-identical for valid values — the round-trip test still passes.

PR #470, opened 2026-07-15
Render DSGo button style variations (primary/secondary/…) at winning specificity

jnealey-godaddy:
Makes `core/button` style variations (`is-style-primary`, `-secondary`, `-header-cta`, and any future one) actually paint on DSGo's single-element buttons, with zero per-variation CSS.

## The problem

DSGo's single-element buttons (icon-button, form submit) don't match WordPress's native `.wp-block-button .wp-block-button__link` descendant rule, so [`class-button-global-styles.php`](../blob/main/includes/features/class-button-global-styles.php) re-emits Global Styles button appearance onto them at specificity **(0,3,0)** (`:root .dsgo-icon-button.wp-block-button__link`).

For variations, `Icon_Button_Styles` already mirrors `core/button.variations.*` onto the icon button and registers them, so WordPress's native pipeline *does* emit variation CSS — but wrapped in `:where()` = **(0,1,0)**, which loses to that (0,3,0) base. So the variation rule was always emitted and always out-ranked; that's why `is-style-secondary` never painted.

## The fix

1. **Project variations at winning specificity.** `Button_Global_Styles` (the generator that already re-emits base button styling) now also iterates `core/button.variations.*` and emits each onto the primitives at a specificity that beats the base rule:
   - Icon Button — `:root .wp-block-designsetgo-icon-button.is-style-{name} .dsgo-icon-button.wp-block-button__link` (0,5,0)
   - Form submit — `:root .dsgo-form__submit.dsgo-form__submit--{name}.wp-element-button` (0,4,0)

   Colours stay defined once in the kit's `core/button.variations`; the slug is strip-sanitised (`[a-z0-9-]`) before it touches the selector. `fill`/`outline` are skipped (base rule + the hand-written `outline` own those).

2. **Remove the now-redundant mirror.** `Icon_Button_Styles::mirror_variation_styles()` (and its two `wp_theme_json_data_*` filters) is deleted — it only ever produced the losing (0,1,0) duplicate. `register_mirrored_variations()` stays, so the variations still appear in the editor Styles panel and are valid to apply.

This makes the generator the single owner of "Global Styles → DSGo button CSS" (base + variations), and leaves exactly one winning rule per variation.

## Scope

- **Icon Button + form submit**: covered. (Form submit's `dsgo-form__submit--{name}` class hooks come from #469; this projection is a harmless no-op until #469 merges, since no element carries the class before then.)
- **Modal Trigger**: intentionally out of scope — it uses its own `buttonStyle` attribute (`dsgo-modal-trigger--{style}`), not block-style variations. The base rule still covers it.

## Validation

Browser-tested on wp-env with a simulated kit (registered `primary`/`secondary`/`header-cta` for `core/button` + their colours in Global Styles):

| Variation | Frontend computed style on `.dsgo-icon-button` |
|---|---|
| `is-style-primary` | filled (kit primary) ✅ |
| `is-style-secondary` | transparent + `2px solid` current-color (ghost) ✅ |
| `is-style-header-cta` | filled (kit accent) ✅ |

- After removing the mirror: the generator's winning rule is present and the native `:where()` duplicate is **gone** (verified by counting matched rules), variations still paint, and the editor panel still lists `primary/secondary/header-cta/fill/outline`.
- `Button_Global_Styles_Variations_Test` (16 tests) pins the sanitiser (incl. selector-breakout attempts), per-primitive selectors, `fill`/`outline` skip, dedup, `:hover`, and empty cases. `phpcs` + `php -l` clean.

## Related

- #469 — Form Builder submit button style variations (adds the `dsgo-form__submit--{name}` hooks this projects onto).

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -25,7 +25,17 @@ class Plugin {
 	 *
 	 * @var string[]
 	 */
-	private const ALLOWED_HOVER_ANIMATIONS = array(
+	/**
+	 * Hover-animation slugs stamped onto buttons as `--{slug}` modifiers.
+	 *
+	 * Public so features that emit button CSS (e.g. Button_Global_Styles) can
+	 * avoid colliding with this class namespace — the form submit uses the same
+	 * `dsgo-form__submit--{x}` modifier space for both animations and style
+	 * variations.
+	 *
+	 * @var string[]
+	 */
+	public const ALLOWED_HOVER_ANIMATIONS = array(
 		'fill-diagonal',
 		'zoom-in',
 		'slide-left',
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -18,18 +18,12 @@ if ( ! defined( 'ABSPATH' ) ) {
  */
 class Plugin {
 	/**
-	 * Allowed hover animation slugs for Icon Button blocks.
-	 *
-	 * Single source of truth used by both the regex check and the
-	 * allowlist validation in apply_default_icon_button_hover().
-	 *
-	 * @var string[]
-	 */
-	/**
 	 * Hover-animation slugs stamped onto buttons as `--{slug}` modifiers.
 	 *
+	 * Single source of truth for the regex check and allowlist validation in
+	 * apply_default_icon_button_hover() / apply_default_form_button_hover().
 	 * Public so features that emit button CSS (e.g. Button_Global_Styles) can
-	 * avoid colliding with this class namespace — the form submit uses the same
+	 * avoid colliding with this namespace — the form submit uses the same
 	 * `dsgo-form__submit--{x}` modifier space for both animations and style
 	 * variations.
 	 *
--- a/includes/features/class-button-global-styles.php
+++ b/includes/features/class-button-global-styles.php
@@ -74,7 +74,8 @@ class Button_Global_Styles {
 	 *   (0,5,0). Matches the hand-written `is-style-outline` rule in
 	 *   `src/blocks/icon-button/style.scss`.
 	 * - Form submit: variation modifier compounded on the button (0,4,0). Matches
-	 *   the `submitButtonVariation` classes in `src/blocks/form-builder/`.
+	 *   the `submitButtonVariation` classes added in #469; until that merges this
+	 *   half is a harmless no-op (no element carries the class yet).
 	 *
 	 * Modal Trigger is intentionally omitted: it uses its own `buttonStyle`
 	 * attribute (`dsgo-modal-trigger--{style}`) rather than block-style
--- a/includes/features/class-button-global-styles.php
+++ b/includes/features/class-button-global-styles.php
@@ -84,8 +84,17 @@ class Button_Global_Styles {
 	 * @var string[]
 	 */
 	private const VARIATION_SELECTOR_TEMPLATES = array(
-		'.wp-block-designsetgo-icon-button.is-style-{name} .dsgo-icon-button.wp-block-button__link',
-		'.dsgo-form__submit.dsgo-form__submit--{name}.wp-element-button',
+		// Icon Button: the variation class sits on the wrapper (`is-style-{name}`)
+		// — a different namespace from the block's `dsgo-icon-button--{anim}`
+		// hover-animation classes — so it can never collide.
+		'icon_button' => '.wp-block-designsetgo-icon-button.is-style-{name} .dsgo-icon-button.wp-block-button__link',
+		// Form submit: the variation class shares the `dsgo-form__submit--{x}`
+		// modifier namespace with hover-animation classes that
+		// apply_default_form_button_hover() stamps on real buttons (e.g.
+		// `dsgo-form__submit--lift`). Emitted only for non-animation slugs — see
+		// build_variation_selector() — so a variation named like an animation can
+		// never repaint a button that merely carries that animation class.
+		'form_submit' => '.dsgo-form__submit.dsgo-form__submit--{name}.wp-element-button',
 	);
 
 	/**
@@ -263,6 +272,13 @@ class Button_Global_Styles {
 	 * base button rule. Colours therefore stay defined once (in the kit /
 	 * theme.json) and are projected onto the icon button / form submit here.
 	 *
+	 * Intentional trade-off: this reads only the `core/button` variation node, not
+	 * a per-block override at `styles.blocks.designsetgo/icon-button.variations.*`
+	 * (which the Site Editor exposes because `Icon_Button_Styles` registers the
+	 * variations for the block). Such a per-block override never rendered anyway —
+	 * WordPress emits it as a `:where()`-wrapped (0,1,0) rule that loses to the
+	 * base button rule — so honouring it is deferred rather than a regression.
+	 *
 	 * @param array $block_styles The core/button block styles from Global Styles.
 	 * @return string Generated CSS, or empty string when there are no variations.
 	 */
@@ -316,8 +332,18 @@ class Button_Global_Styles {
 	 * @return string Selector list, each part prefixed with `:root`.
 	 */
 	private function build_variation_selector( $slug, $pseudo = '' ) {
+		$is_animation_slug = in_array( $slug, Plugin::ALLOWED_HOVER_ANIMATIONS, true );
+
 		$parts = array();
-		foreach ( self::VARIATION_SELECTOR_TEMPLATES as $template ) {
+		foreach ( self::VARIATION_SELECTOR_TEMPLATES as $key => $template ) {
+			// The form-submit modifier namespace is shared with hover animations,
+			// so never emit a form-submit variation rule for an animation slug —
+			// it would repaint any button that merely carries that animation
+			// class. The icon-button selector (wrapper `is-style-{name}`) has no
+			// such collision and is always emitted.
+			if ( 'form_submit' === $key && $is_animation_slug ) {
+				continue;
+			}
 			$parts[] = ':root ' . str_replace( '{name}', $slug, $template ) . $pseudo;
 		}
 		return implode( ",\n", $parts );
--- a/includes/features/class-button-global-styles.php
+++ b/includes/features/class-button-global-styles.php
@@ -271,7 +271,8 @@ class Button_Global_Styles {
 			return '';
 		}
 
-		$css = '';
+		$css  = '';
+		$seen = array();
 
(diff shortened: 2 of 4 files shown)


jnealey-godaddy:
### End-to-end browser validation ✅

Validated on the local wp-env with a simulated Style Kit (registered `primary`/`secondary`/`header-cta` block styles for `core/button` + defined their colours in Global Styles `core/button.variations.*`), then hand-authored icon buttons carrying each `is-style-*` on a published page.

**Generator path (real `wp_get_global_styles`, not reflection):** sees `outline, primary, secondary, header-cta`; emits `is-style-primary` / `-secondary` / `-header-cta`; correctly **skips** `outline`. CSS confirmed injected on the frontend.

**Frontend computed styles on the inner `.dsgo-icon-button`:**

| Variation | Result |
|---|---|
| `is-style-primary` | `background rgb(37,99,235)` (filled) ✅ |
| `is-style-secondary` | `background transparent` + `2px solid` current-color border (ghost) ✅ |
| `is-style-header-cta` | `background rgb(224,101,58)` (filled) ✅ |
| default (no class) | `background rgb(25,25,25)` base fill ✅ |

All three variations paint distinctly, driven entirely by the kit's `core/button.variations` — **zero per-variation CSS in the plugin**, colours defined once in the kit. Screenshot shows blue / ghost / orange / black buttons respectively.

Note: an initial run showed `primary` transparent — that was test data (I'd referenced `var(--wp--preset--color--primary)` with no fallback, and this theme's palette has no `primary` slug); adding a fallback confirmed it paints. Faithful projection of whatever the kit defines.

Still a draft pending the architecture decision (retire hand-written `outline`, modal-trigger mapping, `header-cta` registration) — but the core mechanism is proven end-to-end.

jnealey-godaddy:
Addressed the review in `061e76c` + `c225b84`:

- **Medium (DRY):** extracted a shared `render_rule( $selector, $declarations )` used by both `build_css()` and `build_variation_css()`, removing the duplicated extract→join→wrap pattern. Output is byte-identical (verified against the real theme's Global Styles).
- **Medium (test coverage):** added `Button_Global_Styles_Variations_Test` (16 tests) — pins `sanitize_variation_name()` including selector-breakout attempts (`evil"} body{display:none` → `evilbodydisplaynone`), the per-primitive selectors, the `fill`/`outline` skip, dedup, `:hover`, and the empty cases. Mirrors the `css-escape-bypass-test.php` precedent.
- **Low (numeric slugs):** `sanitize_variation_name()` now casts to string first, so a slug like `2024` (delivered by PHP as an `int` array key) is kept instead of silently dropped.
- **Low (duplicate slugs):** `build_variation_css()` tracks a `$seen` set, so two raw names that normalise to the same slug (`Primary`/`primary`) emit a single rule set.
- **Low (`fill` skip assumption)** and **form-submit no-op (depends on #469):** left as-is — both are the open questions called out in the description for the sync. The `fill`/`outline` skip existing only to avoid double-emitting what the base rule and the hand-written `outline` already own is exactly the "retire the hand-written rule?" decision.

Still a draft pending the architecture call, but the mechanism is now tested and the code is deduped.

jnealey-godaddy:
Great catch on the `Icon_Button_Styles` overlap — this is the most important thread for the sync, so let me lay it out fully.

### Confirmed: the icon-button half overlaps an existing mechanism

`Icon_Button_Styles` (on `main`) already:
- `register_mirrored_variations()` — registers every `core/button` variation as a block style for `designsetgo/icon-button`, and
- `mirror_variation_styles()` — copies `styles.blocks.core/button.variations.*` → `styles.blocks.designsetgo/icon-button.variations.*` on both `wp_theme_json_data_theme`/`_user`.

So WordPress's **native** Global Styles pipeline already emits icon-button variation CSS. The catch is specificity: WP wraps it in `:where()` →
`:root :where(.wp-block-designsetgo-icon-button.is-style-secondary .dsgo-icon-button){…}` = **(0,1,0)**, which loses to `class-button-global-styles.php`'s base rule at **(0,3,0)**. **That specificity loss is exactly why the variations don't paint today** — not a missing rule.

This generator adds `:root .wp-block-designsetgo-icon-button.is-style-secondary .dsgo-icon-button.wp-block-button__link{…}` = **(0,5,0)**, which wins. It works (validated), but it means the icon button now has **two** variation rules — the native (0,1,0) and mine (0,5,0), same values. Redundant, not broken.

### So the real decision for the sync: where does the specificity fix live?

- **(A) This generator as-is** — accept the redundant native rule, or drop the CSS half of `Icon_Button_Styles::mirror_variation_styles()` (keep `register_mirrored_variations()` for the panel) so only the winning rule ships.
- **(B) Fold the high-specificity emit into `Icon_Button_Styles`** (my lean) — icon-button variation logic already lives there, so emit the winning CSS there too, and **reduce this generator to form-submit only**. Form submit is the one primitive that is *not* a block and has *no* native pipeline, so it genuinely needs the projection; icon-button (and eventually modal-trigger) stay with their own block's logic. Cleanest separation, no duplication, one home per primitive.
- **(C) Lower the base rule's specificity** so the native (0,1,0) variation wins — risky: the base is (0,3,0) specifically to beat WP's own element-button rules (documented in `build_css()`), so this could regress the thing it was built to fix.

My recommendation is **(B)**. It turns this PR from "a second, higher-specificity variation emitter" into "form-submit gets what icon-button already has," and keeps `class-button-global-styles.php` focused on the base rule + the non-block primitive.

### The two Low nits

- **Hover emitted without a normal-state rule** (vs `build_css()` gating it): intentional. A variation that defines only `:hover` (e.g. "fill on hover, no base change") *should* still emit its hover rule — matching `build_css()`'s gate would silently drop that intent. Happy to align on strict parity if preferred, but the current behavior is the more-correct one.
- **Sanitiser strips instead of normalising separators** (`header_cta` → `headercta`): the allowlist is deliberately the strict security boundary. For conventional kebab-case block-style slugs — the only real inputs, and kit-controlled — it's identity and matches WP's literal `is-style-{name}` class exactly. `_`/space keys aren't valid block-style slugs. If we want to mirror WP's exact class derivation we'd use `sanitize_title()`, but I kept strict-allowlist as the boundary on purpose.

I'll hold further code changes here until we pick A/B/C at the sync, since (B) would restructure which file owns the icon-button emit.

jnealey-godaddy:
Both addressed in \`6d252f5\`.

**High (form-submit / hover-animation namespace collision):** confirmed live — \`apply_default_form_button_hover()\` stamps \`dsgo-form__submit--{animation}\` on real buttons today, independent of #469, so a \`core/button\` variation named like an animation (e.g. \`lift\`) would emit \`.dsgo-form__submit.dsgo-form__submit--lift {…}\` and repaint any button that merely carries the \`--lift\` hover class. Fixed by skipping the **form-submit** selector for any slug in \`ALLOWED_HOVER_ANIMATIONS\`. The icon-button selector uses the wrapper \`is-style-{name}\` namespace (separate from \`dsgo-icon-button--{anim}\`), so it has no collision and is still emitted. Made \`Plugin::ALLOWED_HOVER_ANIMATIONS\` public so this is one source of truth rather than a fourth copy. Regression test added (\`test_animation_named_variation_skips_form_submit\`): a \`lift\`/\`shrink\` variation emits no \`dsgo-form__submit--lift\` rule but still emits the icon-button rule.

**Medium (per-block override ignored):** intentional, and **not a regression** — a customization at \`styles.blocks.designsetgo/icon-button.variations.*\` never rendered even before this PR (WordPress emits it as a \`:where()\`-wrapped 0,1,0 rule that loses to the 0,3,0 base). Documented the trade-off in \`build_variation_css()\`'s docblock (the class docblock already covered why *writing* the mirror was dropped; this adds why *reading* an override there is deferred). Honouring per-block overrides — merging \`core/button\` + \`designsetgo/icon-button\` variation nodes for the icon-button selector — is a clean follow-up if we want it, but it'd be net-new behavior, not a restoration.

jnealey-godaddy:
All three addressed in \`627ad32\`.

**Medium — incomplete collision guard:** good catch. Widened the guard beyond \`ALLOWED_HOVER_ANIMATIONS\` to the full reserved set for the \`dsgo-form__submit--{x}\` namespace: added \`RESERVED_FORM_SUBMIT_SUFFIXES = ['inline', 'loading', 'no-hover']\` (the layout/state modifiers the block/plugin stamp on real buttons). A variation named \`inline\`/\`loading\`/\`no-hover\` now skips the form-submit selector, same as animation slugs. Test provider extended to cover all three.

**Low — duplicate docblock:** merged the two back-to-back docblocks above \`ALLOWED_HOVER_ANIMATIONS\` into one.

**Low — stale \`register_mirrored_variations()\` wording:** reworded — it no longer claims to make the mirrored node "survive sanitization and render on the front end" (the mirror is gone); it now states registration is for the editor Styles panel + validity, with rendering owned by \`Button_Global_Styles\`.

21 tests green, phpcs clean.

jnealey-godaddy:
Thanks for the approval. The two Lows are addressed in \`4463e03\`:
- **Doc — forward reference to #469:** annotated the form-submit template comment as "added in #469 … no-op until merged."
- **Doc — triplicated docblocks:** trimmed the three near-identical "rendering moved to Button_Global_Styles" explanations to one canonical class docblock + brief pointers in \`init()\` and \`register_mirrored_variations()\`.

**Medium — \`RESERVED_FORM_SUBMIT_SUFFIXES\` is a reactive hand-maintained allowlist (tripwire test):** I'm going to *not* add a source-grepping tripwire, and here's the reasoning. A robust tripwire would have to enumerate every literal \`dsgo-form__submit--{x}\` in the source and assert each is accounted for — but it can't cleanly distinguish a *reserved modifier* (must be in the guard) from a *variation* class like \`--secondary\`/\`--outline\` (must NOT be), so it'd either miss additions or fire false positives, i.e. a fragile test guarding a fragile list.

The real fix for the whole class of bug is to **stop sharing the namespace**: give the form-submit variation its own prefix (e.g. \`is-style-{name}\` on the button, unifying with how the Icon Button already does it). That eliminates \`RESERVED_FORM_SUBMIT_SUFFIXES\`, the skip logic, *and* the collision entirely — no list to keep in sync. It touches #469 (the class hooks) + this template, so it's a focused follow-up rather than something to bolt onto this approved PR. Happy to open it — say the word and I'll rename form-submit variations to a non-colliding namespace across both PRs. Until then, the current guard covers every \`dsgo-form__submit--{x}\` producer that exists in the repo today and is pinned by tests.

PR #472, opened 2026-07-15
refactor(form-builder): namespace submit-button variations under is-style-*

jnealey-godaddy:
## Summary

Follow-up to #469 / #470. The Form Builder submit-button style variation (`secondary` / `outline`) emitted a `dsgo-form__submit--{variation}` class, sharing the BEM modifier namespace with the button's **layout** (`--inline`), **state** (`--loading` / `--no-hover`) and **hover-animation** modifiers. A `core/button` variation named like any of those would collide, so `Button_Global_Styles` carried a `RESERVED_FORM_SUBMIT_SUFFIXES` list plus animation-slug skip logic to suppress the form-submit rule for reserved names.

That guard was fragile: it silently dropped a variation's projected CSS whenever a slug clashed, and its correctness depended on a hand-maintained list staying in sync with three separate modifier sources.

This moves the variation into the **`is-style-{variation}`** namespace — matching how the Icon Button already names its variations — which is disjoint from the `--*` modifiers. The collision class disappears entirely.

## What changed

- **`save.js` / `edit.js`** — submit button now emits `is-style-{variation}` (validated against the block.json enum, unchanged).
- **`style.scss`** — ghost fallback selectors retargeted to `.dsgo-form__submit.is-style-{secondary,outline}.wp-element-button` (specificity preserved at 0,4,0). Refreshed the now-stale hover-guard comment.
- **`class-button-global-styles.php`** — `form_submit` template → `is-style-{name}`; **deleted `RESERVED_FORM_SUBMIT_SUFFIXES` and the reserved-slug skip branch**. The generator now projects every registered `core/button` variation onto the submit button unconditionally.
- **`class-block-inserter.php`** — AI-inserted forms emit the same `is-style-{variation}` class, matching `save()`.
- **Tests** — the three affected PHPUnit suites updated in lockstep; the old `test_reserved_slug_variation_skips_form_submit` is replaced by `test_modifier_named_slug_stays_in_is_style_namespace`, which proves a slug like `lift`/`inline`/`loading` now emits safely in the `is-style-*` namespace and never in `--*`.

## No deprecation needed

`submitButtonVariation` shipped in **no released version** — it lives in the `[Unreleased]` #469 changelog entry, added after the 2.4.0 tag. No published content carries the old class, so this is pre-release refactoring, not a schema migration. The block still round-trips its own `save()` output.

## Verification

- `npm run build` — clean; `build/blocks/form-builder/index.css` carries `is-style-*`, no `--secondary`/`--outline`.
- `lint:js` / `lint:css` / `phpcs` — clean on changed files.
- **JS round-trip: 218/218** (`deprecations-isEligible.test.js`) — form-builder parses its own current `save()` as valid and round-trips losslessly.
- **PHPUnit: 29/29** across `Button_Global_Styles_Variations_Test`, `Block_Inserter_Form_Variation_Test`, `Form_Builder_Hover_Guard_Sync_Test` (WP 6.9.4 / PHP 8.3).


jnealey-godaddy:
Thanks for the review. Addressed the one Low finding and validated the change in a real browser.

**Low — stale `ALLOWED_HOVER_ANIMATIONS` docblock (`includes/class-plugin.php`)** — fixed in `03cadb3a`. The note claimed the form submit shares the `dsgo-form__submit--{x}` space for both animations and style variations; it now states variations live in the disjoint `is-style-{x}` namespace, so a future reader won't re-derive the deleted reserved-slug logic from it.

**Browser validation (Chrome DevTools MCP)** — published a page with three form-builder blocks (default / secondary / outline) and read computed styles on the live frontend (WP 6.9.4, Twenty Twenty-Five):

| button | class | background | border |
|---|---|---|---|
| default | `dsgo-form__submit … --fill-diagonal` | `rgb(25,25,25)` (filled) | 0px |
| secondary | `dsgo-form__submit is-style-secondary … --fill-diagonal` | `rgba(0,0,0,0)` (ghost) | 1px solid currentColor |
| outline | `dsgo-form__submit is-style-outline … --fill-diagonal` | `rgba(0,0,0,0)` (ghost) | 2px solid currentColor |

Note the secondary/outline buttons **also carry `dsgo-form__submit--fill-diagonal`** (the theme's default hover animation, auto-stamped by `apply_default_form_button_hover()`) and still paint their variation correctly — the exact both-namespaces-on-one-element case the refactor is meant to make safe. Both the style.scss ghost rule and the `Button_Global_Styles` `is-style-*` projection were confirmed present in the live stylesheet. Test page and artifacts cleaned up afterward.

PR #482, opened 2026-07-21
fix: source block labels from HTML so translation doesn't invalidate blocks

jnealey-godaddy:
## Problem

Translating a site (via the site-designer flow) put **"Block contains unexpected or invalid content" → Attempt recovery** on DSGo button/label blocks.

Root cause: those blocks stored their visible label as a **plain block-comment JSON attribute** *and* rendered the same value into the HTML — two sources of truth. The translation path rewrites the visible HTML text but leaves the comment JSON stale, so `save()` regenerates the old label, it no longer matches the stored markup, and the block is flagged invalid. Core buttons are immune because core sources its label from the HTML only.

## Fix

Source each label from the HTML (`source: 'html'` for `RichText.Content` renders, `source: 'text'` for plain-text renders), matching `core/button`'s single-source-of-truth model.

`save()` markup is **byte-identical**, so existing content stays valid with **no deprecation**, and because the sourced value overrides the stale comment, already-broken translated content is **healed** on update.

### Scope of this PR — only "always-present label" blocks

Limited to blocks whose label element is always in the markup when the text is set (rendered unconditionally, or gated solely on the text's own presence, `{title && …}`). For those, an absent element unambiguously means empty text, so a sourced value can never silently diverge from what the author typed.

| Block | Attribute(s) | Source |
|---|---|---|
| icon-button | `text` | html |
| modal-trigger | `text` | html |
| accordion-item | `title` | html |
| timeline-item | `title`, `date` | html |
| counter | `label` | text |

## Deferred to a consolidated deprecation-based follow-up

These four blocks need a `save.js` markup change **and a deprecation**, so they're intentionally out of scope here:

- **card** (`title`/`subtitle`/`bodyText`/`badgeText`) and **table-of-contents** (`titleText`) render their label **only when a separate `show*` toggle is on**. Sourcing them (as an earlier revision of this PR did) causes silent data loss: toggle visibility off → the element disappears → the sourced value resets to default on reparse. *(Caught in review — reverted in 3f7d426.)* The fix is to always render the element (hidden via class) plus a deprecation that preserves existing hidden-but-filled content.
- **form-builder** (`submitButtonText`) and **countdown-timer** (`completionMessage`) duplicate the label into a wrapper `data-*` attribute AND the visible text; a partial translation desyncs the two copies. The redundant `data-*` copy must be removed — again a markup change requiring a deprecation.

## Verification

- Full unit suite green, plus new `tests/unit/translation-resilience.test.js` pinning both invariants (translation stays valid; legacy stale-comment content heals) across the in-scope attributes. The file documents exactly why each excluded block is unsafe to source without the follow-up.
- Build + lint clean; pre-commit e2e passed.
- **Live editor check** (earlier revision): a page with stale-English comment + translated HTML rendered valid with no Attempt-recovery banner.

## Also included

`649910b8` carries over two unrelated uncommitted working-tree style fixes: grid item `overflow-wrap: anywhere`, and Pills-variation tab hover text following the active/accent color.

### DIFF
--- a/src/blocks/accordion-item/block.json
+++ b/src/blocks/accordion-item/block.json
@@ -20,6 +20,8 @@
 	"attributes": {
 		"title": {
 			"type": "string",
+			"source": "html",
+			"selector": ".dsgo-accordion-item__title",
 			"default": "Accordion Item"
 		},
 		"isOpen": {
--- a/src/blocks/card/block.json
+++ b/src/blocks/card/block.json
@@ -129,8 +129,6 @@
 		},
 		"badgeText": {
 			"type": "string",
-			"source": "text",
-			"selector": ".dsgo-card__badge",
 			"default": ""
 		},
 		"badgeStyle": {
@@ -169,20 +167,14 @@
 		},
 		"title": {
 			"type": "string",
-			"source": "html",
-			"selector": ".dsgo-card__title",
 			"default": ""
 		},
 		"subtitle": {
 			"type": "string",
-			"source": "html",
-			"selector": ".dsgo-card__subtitle",
 			"default": ""
 		},
 		"bodyText": {
 			"type": "string",
-			"source": "html",
-			"selector": ".dsgo-card__body",
 			"default": ""
 		},
 		"overlayOpacity": {
--- a/src/blocks/card/block.json
+++ b/src/blocks/card/block.json
@@ -129,6 +129,8 @@
 		},
 		"badgeText": {
 			"type": "string",
+			"source": "text",
+			"selector": ".dsgo-card__badge",
 			"default": ""
 		},
 		"badgeStyle": {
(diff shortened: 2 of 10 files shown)


jnealey-godaddy:
Good catch — verified and confirmed. `card` (`title`/`subtitle`/`bodyText`/`badgeText`) and `table-of-contents` (`titleText`) render their label only when a separate `show*` toggle is on, so sourcing them resets the value to default when the element is toggled out of the markup. This hits both new content and existing hidden-but-filled fields.

One refinement to the suggested fix: always-rendering the element (hidden via class) also needs a **deprecation** — existing `show:false` cards have no element and store the text in the comment, so without a deprecation that reads it, the text is still lost (silently, just without the recovery banner).

Resolution (3f7d426): reverted sourcing for those five attributes. This PR now covers only "always-present label" blocks (icon-button, modal-trigger, accordion-item, timeline-item, counter) — no markup change, no deprecation, no regression. card/TOC move into one consolidated follow-up alongside form-builder (`data-submit-text`) and countdown-timer (`data-completion-message`), which all need the same always-render + deprecation treatment. The test file now documents why each excluded block is unsafe to source without that work.

The Medium test-gap note is addressed by scope: the remaining in-scope blocks have no `show*` toggle, so the toggle-off case can't apply to them; the toggle round-trip test will land with the card/TOC fix in the follow-up.

PR #483, opened 2026-07-21
fix: source labels for the deferred blocks (form-builder, countdown-timer, table-of-contents, card)

jnealey-godaddy:
## Context

Follow-up to #482, which fixed translation-invalidation for the "always-present label" blocks by adding `source` to `block.json` with no markup change. This PR handles the four blocks that were deferred because they need a `save.js` markup change **and a deprecation**. They split into two shapes.

## Shape A — label duplicated into a `data-*` attribute (plain markup deprecation)

**form-builder** (`submitButtonText`) and **countdown-timer** (`completionMessage`) each stored their label twice in the markup: once as the visible text and once as a `data-*` attribute on the wrapper. Translating the visible text left the data attribute stale → `save()` no longer matched → Attempt recovery.

- **form-builder**: `data-submit-text` is unused on the frontend (view.js restores the label from `button.textContent`). Removed it; `submitButtonText` now sources from `.dsgo-form__submit`.
- **countdown-timer**: the message div was CSS-hidden and its text overwritten by JS from `data-completion-message`. Kept the message as the div's text (still translatable), sourced `completionMessage` from it, removed the duplicate `data-completion-message`, and simplified view.js to just reveal the (already server-rendered) div.

Each gets a markup-change deprecation (`apiVersion: 3`) reproducing the old markup, so existing content stays valid and re-serializes without the redundant attribute.

## Shape B — label rendered only when a `show*` toggle is on (data-preserving deprecation)

**table-of-contents** (`titleText`) and **card** (`title`/`subtitle`/`bodyText`/`badgeText`) rendered their label only when a separate toggle was on. Sourcing from an element that can disappear would reset the text whenever the field was hidden — the data-loss regression caught in #482's review. Fix: always render the element when the text is set and hide it with a `--hidden` class (CSS `display:none`) when the toggle is off, so the sourced value survives visibility toggles.

- **table-of-contents**: the old gate was `{showTitle && ...}` (not text-dependent), so always-render reproduces shown-title markup exactly. A plain **markup-change** deprecation (v1) migrates content saved with the title hidden.
- **card**: the old gates were `{showTitle && title && ...}` (text-dependent) with empty defaults, so hidden-field content stays *valid* under the new save() (empty source → nothing renders) and a markup deprecation would never fire. Uses an **isEligible** deprecation (v1) that opts those valid blocks into migration when a field's comment text is present but its element is missing, preserving the author's hidden text. (card had no `deprecated.js`; created + wired into index.js.)

## Verification

- Full unit suite green (**4984 tests**), including two new regression suites:
  - `tests/unit/label-dedup-deprecation.test.js` — translation-safety + legacy migration for form-builder/countdown-timer.
  - `tests/unit/conditional-visibility-deprecation.test.js` — translation-safety, **toggle-off preservation** (the core regression), and legacy migration for TOC/card.
- `deprecations-isEligible.test.js` passes — no deprecation reclaims current content; lossless round trips.
- Build + JS/CSS lint clean; pre-commit e2e green.
- **Live editor check** (real WP): a page with legacy markup for all four blocks (stale `data-*` attributes; hidden-title/subtitle with text in the comment) loads with **zero "Attempt recovery"** — form-builder and countdown-timer migrate off their redundant attributes, TOC/card preserve their hidden text.

With this + #482, the whole translation-invalidation class is resolved across all affected blocks.

Plan: `docs/plans/2026-07-21-conditional-and-dup-label-sourcing.md`.

## Regenerated patterns — what changed and why (17 regions, 19 files)

Ran `tools/regenerate-patterns.js` for form-builder, countdown-timer, and card (table-of-contents patterns needed no changes). The diffs are mechanical and **visually identical** — no design change. Three kinds of change appear:

1. **Label-sourcing (this PR):** `data-submit-text` / `data-completion-message` removed from the wrapper; `submitButtonText`/`title`/`subtitle`/`bodyText` removed from the block comment (now DOM-sourced). The visible text stays in the button / message div / heading, so output is unchanged.
2. **`--dsgo-form-border-color:#d1d5db` (+ empty `--dsgo-form-label-color:` / `--dsgo-form-field-bg:`) removed from inline `style`:** fallout from re-running the generator, which also applies the older v4/v5 deprecations. Not a regression — `.dsgo-form-builder` in `style.scss` supplies `--dsgo-form-border-color: #d1d5db` as the CSS default (verified in the built CSS), so the border is byte-for-byte the same colour; the other two custom properties were empty and had no effect.
3. **Form-builder sizing attributes (`fieldSpacing`/`inputHeight`/…) now explicit in the comment:** the v5 migration promotes the values that were baked into the inline `style` into named attributes. The `style` attribute still carries the same `--dsgo-form-field-spacing:1.5rem` etc., so spacing/sizing is unchanged.

In short: reviewers can treat the 16 near-identical form/countdown diffs as "strip the now-redundant duplicate label copies; inherit the border colour from CSS instead of inline." No visual diff.


jnealey-godaddy:
Thanks — verified each finding against the code before acting. Pushed f3eae62.

**🟡 Medium — bundled patterns not regenerated: confirmed and fixed.** Ran `tools/regenerate-patterns.js` for both blocks:
- **form-builder** (12 patterns) — drops `data-submit-text` and moves `submitButtonText` out of the block comment (now sourced from the button). The regeneration also cleaned up the pre-existing stale `--dsgo-form-border-color:#d1d5db` fallback from the v4-era change.
- **countdown-timer** (5 patterns) — drops `data-completion-message`; the message stays as the div's text.

One correction to the review's note that card/TOC patterns are unaffected: **card patterns *were* affected** — the dry run flagged 9 regions across 2 files (`pricing-cards`, `features-grid`). They weren't *invalid* (title sourced from the `<h3>` matched, so no Attempt Recovery), but they still carried `title`/`subtitle`/`bodyText` as redundant duplicates in the block comment. Regenerated those too — HTML output unchanged, comment-only diff. **table-of-contents** genuinely needed 0 changes (0 regions). `grep` now finds zero `data-submit-text` / `data-completion-message` in `patterns/`, and `php -l` passes on all 19 files.

**🟢 Low — clsx/classnames: fixed.** Swapped `card/save.js` to `classnames` (`clsx` was only in save.js, not deprecated.js). `classnames` is a direct dependency (`^2.3.2`); `clsx` was transitive-only, so this also removes the fragile hoisted resolution. Class output is byte-identical — the `deprecations-isEligible` "parses its own save() as valid" check and all 243 deprecation/regression tests still pass.

**🟢 Low — CHANGELOG: intentionally deferred.** The working tree already has in-progress release-prep on `CHANGELOG.md` (plus `package.json` / `designsetgo.php` / `readme.txt` version bumps) that predates this branch — folding `[Unreleased]` into the 2.5.0 section. I didn't want to entangle that unrelated release work into this PR, so I left it untouched rather than commit a mixed changelog. Happy to add the entry (covering #482 + #483) once the release-prep lands, or into whichever changeset owns it.

jnealey-godaddy:
Thanks for the approval — both Low nitpicks addressed in 8571a1b:

- **Plan-doc drift** — corrected the countdown-timer section to describe the shipped design (`completionMessage` sourced via `source: 'text'` from the message div, single source of truth that stays translatable), and noted why the earlier `source: 'attribute'` draft was dropped (WordPress doesn't resolve a root-wrapper attribute source there).
- **Substring matching in `card/deprecated.js`** — the isEligible guard now matches the field class as a whole class token (quote/whitespace-delimited) via a small `hasClassToken()` helper, instead of `html.includes()`, so a future sibling class that shares a prefix can't be mistaken for the field's element. Behavior is identical for the current class names; the 15 conditional-visibility tests and the full `deprecations-isEligible` suite still pass. Left the pre-existing `countdown-timer` v3 guard as-is (out of scope, and works).

jnealey-godaddy:
Both addressed in 2d32c6c — and the second one turned out to be more than cosmetic:

- **Medium — stale v4 docblock (countdown-timer):** confirmed and fixed. The JSDoc header was leftover from the first `source: 'attribute'` draft and contradicted the code beneath it. Rewrote it to match what shipped (message sourced from the div text; `data-completion-message` is the redundant copy that's dropped).

- **Low — `hasClassToken` false-negative (card):** verified this is a real (if unlikely) gap and closed it. The check now scopes to `class="..."` attribute values — `class="(?:[^"]*\s)?${className}(?=[\s"])` — so a field's own RichText body copy that literally contains a sibling class-name token (e.g. "…the dsgo-card__title class…") can't be mistaken for that element being present, which would otherwise return `isEligible: false` and silently skip migrating the hidden field. Added a regression test (`still migrates a hidden title when visible body text mentions the title class name`) that reproduces exactly that: a hidden title whose body text contains `dsgo-card__title`. I checked it fails against the previous whole-markup regex and passes now, so it pins the fix rather than just documenting it. Full suite is 244 green.

Left the pre-existing `countdown-timer` v3/v5 `html.includes()` guards untouched (out of scope for this PR, and not reachable by the same body-text vector since they key on structural classes/tokens rather than author text).

jnealey-godaddy:
Good catch — this was a real gap I missed, and verifying it live turned up more than the diff alone showed. Fixed in 182022e8.

**Fixed (the label-sourcing parity the finding flagged):** updated `class-block-inserter.php` to mirror each `save.js` change:
- countdown-timer: no longer emits `data-completion-message` (message is the div text).
- form-builder: no longer emits `data-submit-text` (label is the button text).
- card: renders `title`/`subtitle`/`bodyText` whenever the text is set, adding `--hidden` when the toggle is off (was `if ($show_* && $text)`, which dropped the element — and the DOM-sourced text — when hidden).
- table-of-contents: always renders the title element with `--hidden` when off.

Added `tests/phpunit/block-inserter-label-sourcing-test.php` (6 tests) pinning all four, and confirmed the existing 31 `Block_Inserter` tests still pass.

**Verified live** by generating each block through `Block_Inserter::build_block_markup()` (hidden-field scenarios) and opening the result in the editor:
- **card** ✅ and **countdown-timer** ✅ now validate clean — no Attempt recovery, and the card's hidden title text is preserved (sourced from the now-present `--hidden` `<h3>`).

**What the live test also surfaced — pre-existing inserter drift, unrelated to this PR (flagging, not fixing here):** form-builder and table-of-contents still show Attempt recovery, but *not* from the label-sourcing — the console diffs show the remaining mismatches are older inserter/save drift that exists on `main` and is out of scope for a translation fix:
- **form-builder:** the inserter still bakes `data-email-to`/`data-enable-email`/… (7 attrs), the nullable spacing tokens (`--dsgo-form-field-spacing` etc.), and inline submit-button sizing — none of which `save.js` emits. Notably `main`'s `form-builder/save.js` emits **zero** `data-email-*` (there's already a deprecation for the version that did), so the inserter is stale from that earlier save() change, not this one.
- **table-of-contents:** the only remaining diff is `data-unique-id` (the inserter generates a value into the markup but not the block comment, so it doesn't round-trip).

With the label-sourcing noise removed, the form-builder markup *does* match save() — so this PR's change is correct and complete for label-sourcing. I'd rather not fold the email-attr/spacing/`uniqueId` cleanup into a translation PR (removing the `data-email-*` attrs in particular deserves a check against the email feature before ripping them out), so I'll open a separate focused "inserter ↔ save() parity" pass for form-builder + TOC. Happy to prioritize that next if you'd prefer it bundled here instead.

jnealey-godaddy:
**Medium — pattern regen scope: confirmed no visual regression, and documented in the PR body.**

The 16 near-identical diffs are all mechanical. Two categories beyond label-sourcing, both visually identical:
- `--dsgo-form-border-color:#d1d5db` dropped from inline `style` → not a regression: `.dsgo-form-builder` in `style.scss` (L26) supplies `--dsgo-form-border-color: #d1d5db` as the CSS default (verified it's in the built CSS), so the input border renders the exact same colour. The sibling `--dsgo-form-label-color:` / `--dsgo-form-field-bg:` that also disappear were empty and had no effect.
- The new `fieldSpacing`/`inputHeight`/… comment attributes are the v5 deprecation promoting values that were baked into inline `style` into named attributes — the `style` still carries the same `--dsgo-form-field-spacing:1.5rem` etc., so spacing/sizing is unchanged.

I added a **"Regenerated patterns — what changed and why"** section to the PR description that categorizes all three kinds of change, so reviewers don't have to reverse-engineer the diffs. Net: strip the redundant duplicate label copies + inherit the border colour from CSS instead of inline; no design change.

**Low — card badge not rendered by the inserter: confirmed pre-existing, and lower-exposure than it looks.**

I checked `includes/abilities/` and `badgeText`/`showBadge` aren't in the card ability's input schema at all — so there's no path to *pass* them through the AI ability today; the drop would only happen if something called `build_block_markup()` with a badge attribute directly. Adding faithful badge rendering to the inserter (floating vs inline positions, position/colour modifiers, the new `--hidden` class) plus exposing the params in the schema is a small feature addition rather than a parity fix, so I'll fold it into the same focused inserter-parity follow-up as the form-builder `data-email-*` / TOC `uniqueId` drift rather than expand this translation PR. It's valid (not silent-loss) in the meantime because the card inserter simply never emits a badge, which matches `save('')`.

PR #484, opened 2026-07-21
fix: compatibility deprecations for site-designer-generated form-builder + grid

jnealey-godaddy:
## Problem

Pages built by the Airo site-designer API show "Attempt Recovery" on some blocks after updating to the plugin's current `save()`. The generator emits markup that never exactly matched *any* plugin version's `save()`, so no existing deprecation reproduces it and the block can't silently migrate. This heals that deployed content with two targeted "compatibility" deprecations — reproduce the generator's exact shape, then `migrate()` to clean current markup.

Diagnosed from a real generated page (post 108): `form-builder` and one `grid` were invalid; the other three grids on the page were already valid.

## Form Builder — `v7`

The generator emits a **v4-era** form shape (baked `--dsgo-form-*` tokens, inline submit sizing, a `data-submit-text` copy of the label) but its serializer **strips the honeypot's `aria-hidden` and the message div's `aria-atomic`** — attributes present in *every* real plugin version's `save()` and every existing deprecation, so nothing matched.

`v7` reproduces that exact shape (v4 `save()` **minus** the two ARIA attributes). `successMessage`/`errorMessage` are **sourced from the wrapper's `data-*` attributes**, because the generator writes a custom message into the HTML without mirroring it into the block comment. `migrate()` strips the baked default tokens so the form inherits the theme.

Result on migration: ARIA restored, `data-submit-text` + baked tokens dropped, **custom error message preserved**. Real plugin content carries the ARIA attributes and never matches this `save()`.

## Grid — `legacyResponsiveTabletClass`

The generator emits grids whose **tablet column count lives in a `className`** (`dsgo-grid-cols-tablet-1`) while the comment's `tabletColumns` drifted to the default (`2`), and whose **min-width lives only in the inner CSS** (`minmax(480px, 1fr)`) with no `columnMinWidth` attribute. The current `save()` emits a *second*, attribute-derived tablet class (`dsgo-grid-cols-tablet-2`) the stored markup never had → invalid.

Key enabler: **WordPress compares the `class` attribute as an unordered set**, and the existing `legacyMinWidth` deprecation already reproduces the inner track verbatim (via `legacyInnerStyle`). So the new deprecation reproduces the stored class set by **omitting the attribute-derived tablet class** (the stored one comes from `className`), and `migrate()` recovers `columnMinWidth` from the track, lifts the real `tabletColumns` out of the class, and drops the redundant class — **preserving the design** (tablet-1, 480px min-width).

## Verification

- New regression tests built from the **exact markup on the page** (`form-builder-compat-deprecation.test.js`, `grid-compat-deprecation.test.js`).
- `deprecations-isEligible.test.js` green — neither compat entry reclaims current content (both are markup-change deprecations, reached only by save-matching on the already-invalid generated HTML; real plugin content never matches).
- Updated `grid/test/deprecated.test.js` for the new array order.
- Build + lint clean; pre-commit e2e green.
- **Live-verified** in the editor: both migrate silently (`Block successfully updated`, no Attempt Recovery). Form — ARIA restored, custom error preserved. Grid — `tabletColumns:1`, `columnMinWidth:480px`.

## Scope / follow-up

These cover the two shapes actually present on the diagnosed page. The generator may emit other variants across its catalog; capturing failing markup from a few more generated sites (or the `block-patterns` validation harness) would surface any remaining shapes, each fixable the same way. The api is separately being corrected to emit `save()`-matching markup going forward; these deprecations heal already-deployed content on any version.


jnealey-godaddy:
Addressed the review notes in df7bb5d3:

**v7/v2 ordering overlap (the one flagged as worth confirming).** Investigated empirically rather than by inspection. The feared shadowing does not occur: v7 and v2 resolve on *different* `save()` output. v7 sources the success/error messages from the wrapper's `data-*` attributes (the generator's HTML-only shape), while v2 sources `errorMessage` from the block comment. So:

- Generator content (custom error only in `data-error-message`, not in the comment) → **v7** claims it and strips the baked spacing tokens.
- Genuinely-old v2-era content (error message mirrored into the comment) → still **v2**, whose passthrough `migrate()` preserves the baked tokens.

Confirmed by tagging each `migrate()` and parsing both shapes through the real `@wordpress/blocks` parser: byte-identical HTML bodies route to different deprecations purely on the comment's `errorMessage` key. Pinned this permanently with a regression test (`does not let v7 shadow genuinely-old v2-era content`) that asserts the v2-era shape stays valid and keeps its baked `fieldSpacing`, i.e. v7 did not re-claim it.

**Missing CHANGELOG entry.** Added to both CHANGELOG.md (technical) and readme.txt (plain-language) under 2.5.0, in the same 'Attempt Recovery' family as the translation-recovery fix.

**Selector-sourcing note.** Added the inline comment explaining why `selector: '[data-success-message]'` reaches the wrapper's own attribute (hpq matches against a body whose innerHTML is the block markup).

Left the v7/v4 duplication as-is on purpose — it's the file's deliberate frozen-snapshot pattern, and a shared helper would couple the deprecations together.

PR #491, opened 2026-07-27
fix(overlay-header): measure the header, not the footer, and clear the hero

jnealey-godaddy:
## Description

The overlay header was unusable on Twenty Twenty-Five: the header stayed opaque grey instead of going transparent, and the hero was yanked up far past the header, slicing off its heading. Two independent bugs, plus a spacing gap that made the result look cramped even once both were fixed.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement

## Changes Made

**1. The content pull-up was sized from the footer, not the header.**

`setupOverlayHeaderHeight()` ran for every element matched by the sticky-header selector, which includes `.wp-block-template-part:has(.wp-block-navigation)` — that matches the footer on most themes. Each match wrote the shared `--dsgo-overlay-header-height` custom property, so the last one in DOM order won:

| Element | Height |
| --- | --- |
| `<header>` (correct) | 92px |
| `<footer>` (what won) | **410px** |

Content got `margin-top: -410px` instead of `-92px`. The measurement is now restricted to the single template part that is a direct child of `.wp-site-blocks`, which is what the CSS pull-up actually consumes.

**2. The header never went transparent.**

The strip rule only matched `.has-background`. A theme can paint a container background from a block style variation — Twenty Twenty-Five's `is-style-header-section` — or from `styles.blocks` in `theme.json`, neither of which adds that class. The rule now matches container block classes as well.

Backgrounds are deliberately preserved on buttons, submenu dropdowns, and the mobile menu overlay; all three are unreadable over a hero image without one.

**3. Hero clearance.**

The header height is added to the first content section's top padding, so the hero's content clears the header while its background still runs behind it. The authored padding is kept as a CSS string rather than a pixel snapshot, so fluid presets keep responding to viewport changes:

```
calc(var(--wp--preset--spacing--60) + var(--dsgo-overlay-header-height, 0px))
```

Both terms stay live, so a resize needs no re-run. The authored value is cached in a data attribute so repeat runs can't nest the `calc()` inside itself and compound the clearance.

## Testing

- [x] Tested in WordPress editor
- [x] Tested on frontend
- [x] Tested with Twenty Twenty-Five theme
- [x] Tested responsive behavior (mobile/tablet/desktop)
- [x] No console errors
- [x] No PHP errors

Verified against a live Twenty Twenty-Five site with the compiled CSS and the new JS logic:

| Check | Result |
| --- | --- |
| Overlay height source | 92px header, no longer the 410px footer |
| Header at rest | transparent |
| Header scrolled | `rgba(255, 255, 255, 0.95)` fades in, nav readable |
| Hero clearance (1512px) | 196px = 104 authored + 92 header |
| Hero clearance (390px) | 176px = 104 authored + 72 header, no compounding |
| CTA button | keeps its background |
| Mobile menu overlay | keeps its background |

4615 unit tests and the e2e suite pass. `npm run build`, `lint:js`, and `lint:style` are clean.

## Additional Notes

No PHP changes, no attribute or markup changes, so no deprecation is needed. The `:not(<complex-selector>)` forms used in the exception list are supported across all browsers in the project's support range.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Overlay header is transparent again, and no longer pulls page content up by the footer's height.** Two independent bugs made the per-page Overlay Header unusable on Twenty Twenty-Five. The content pull-up was sized from the wrong element: `setupOverlayHeaderHeight()` ran for every element matched by the sticky-header selector, which includes `.wp-block-template-part:has(.wp-block-navigation)` — that matches the footer on most themes — and each match wrote the shared `--dsgo-overlay-header-height` custom property, so the last writer in DOM order won and content was pulled up by the footer's height (410px) instead of the header's (92px), slicing the top off the hero. Measurement is now restricted to the single template part that is a direct child of `.wp-site-blocks`, which is what the CSS pull-up actually consumes. Separately, the header never went transparent, because the background-strip rule only matched `.has-background`: a theme can paint a container from a block style variation (Twenty Twenty-Five's `is-style-header-section`) or from `styles.blocks` in `theme.json`, neither of which adds that class. The rule now matches container block classes as well, with a zero-specificity `:where()` exclusion list that preserves backgrounds on buttons, submenu dropdowns, and the mobile menu overlay — all unreadable over a hero image without one. The first content section also gains the header height as top padding, so its content clears the header while its background still runs behind it; the authored padding is preserved as a CSS string rather than a pixel snapshot, so inline fluid presets keep responding to viewport changes, and `--dsgo-overlay-hero-clearance: 0px` opts a section out. (#491)
+
 ## [2.5.1] - 2026-07-23
 
 ### New Features
--- a/jest.config.js
+++ b/jest.config.js
@@ -39,7 +39,8 @@ module.exports = {
 
 	// Module name mapper for CSS and asset files
 	moduleNameMapper: {
-		'\\.(css|less|scss|sass)$': 'identity-obj-proxy',
+		'\\.(css|less|scss|sass)$':
+			'<rootDir>/tests/unit/__mocks__/styleMock.js',
 		'\\.(jpg|jpeg|png|gif|svg|eot|otf|webp|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
 			'<rootDir>/tests/unit/__mocks__/fileMock.js',
 		// @wordpress/editor and @wordpress/notices pull in heavy WP store internals
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "designsetgo",
-  "version": "2.5.0",
+  "version": "2.5.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "designsetgo",
-      "version": "2.5.0",
+      "version": "2.5.1",
       "hasInstallScript": true,
       "license": "GPL-2.0-or-later",
       "dependencies": {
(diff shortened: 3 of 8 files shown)


jnealey-godaddy:
Thanks — worked through all of it. Four changes, one respectful disagreement.

**Medium — hero padding has no opt-out.** Added one: `--dsgo-overlay-hero-clearance: 0px` on the block.

I did keep the padding unconditional rather than sniffing for a hero-shaped block, because the two aren't independent. The CSS pull-up already applied `margin-top: calc(-1 * var(--dsgo-overlay-header-height))` to that exact element; the padding is its counterpart on the same element, so content lands back where it started. Skipping it for "non-hero" first blocks wouldn't avoid unwanted padding — it would leave that block's content occluded by the header. The opt-out is the right escape hatch, so that's what it now has.

**Low — fluid-padding comment is wrong for class-sourced padding.** Correct, and the comment overstated it. Now says the string only stays fluid when the block serialized its padding inline (how WordPress writes block spacing); stylesheet-inherited padding has no inline string to keep and falls back to a pixel snapshot taken at init.

**Low — specificity.** Agreed, wrapped the exclusion chain in `:where()`. Six chained `:not()`s were adding roughly six class-selectors' worth over the `.has-background` rule they replace. `!important` is what does the overriding here, so the chain should contribute nothing. Verified in-browser afterwards that the header still goes transparent and the CTA button still keeps its background.

**Low — `getOverlayTarget()` ignores `customSelector`.** Declining this one, and I think the premise doesn't hold. The CSS pull-up that consumes this variable is itself hardcoded to `.wp-site-blocks >`:

```css
body…dsgo-page-overlay-header .wp-site-blocks > header.wp-block-template-part + *,
body…dsgo-page-overlay-header .wp-site-blocks > .wp-block-template-part:not(footer):first-of-type + *
```

Overlay has never honored `customSelector`. If a custom selector matched a header elsewhere in the tree, measuring it would write a variable that no rule reads. The JS mirroring the CSS is the correct coupling; honoring `customSelector` here would reintroduce a second writer, which is the bug this PR fixes. Happy to revisit if the pull-up itself is generalized.

**Informational — no JS test coverage.** Agreed, this was the real gap. Added `tests/unit/overlay-header-measurement.test.js`: jsdom builds a header *and* a footer that both contain a navigation, then asserts the variable tracks the header. Removing the guard makes it report `410px` instead of `92px`, so it pins this exact regression rather than just passing. Also covers the clearance calc, no-compounding across a soft navigation, and the non-overlay no-op.

That surfaced a latent config bug: `jest.config.js` mapped stylesheets to `identity-obj-proxy`, which isn't installed — no existing test imports a module that imports its own SCSS, so it had never been exercised. Pointed it at a local `styleMock.js`, matching the existing `fileMock.js` treatment of asset imports.

**Informational — CHANGELOG.** Added under `[Unreleased] → Fixed`.

**Separately: Security Audit.** Was failing on GHSA-mv8w-475r-vwqw (`seroval` type confusion in `fromJSON()`, vulnerable `<=1.5.2`). Nothing depends on it directly — it resolved to 1.5.0 six levels down via `@wordpress/editor` → `@wordpress/interface` → `@wordpress/admin-ui` → `@wordpress/route` → `@tanstack/react-router` → `@tanstack/router-core`. Pinned to `^1.5.6` through the existing `overrides` block; it satisfies `seroval-plugins`' `^1.0` peer so the tree dedupes to one copy. `npm audit --audit-level=critical` passes now.

This failure predated the branch and reproduced on a clean checkout of `main` — flagging it since it'll affect other open PRs until it lands somewhere.

Local verification: 159 suites / 4619 tests, build, both linters, and the e2e suite all pass.

jnealey-godaddy:
Thanks for the second pass. Took the comment you asked for on the Medium; explaining rather than changing the Low.

**Medium — DOM-depth assumption in `applyOverlayHeroPadding()`.** Added the defensive comment (a05dbf36). It records why the clearance lands one level in rather than on the pulled-up element: padding the pulled-up element itself would push its whole box back down and cancel the overlay, while padding the first block inside it moves only the content and leaves that block's background under the header.

One correction on the failure mode, though — an extra wrapper level doesn't reproduce the "content sliced off" symptom. That bug was an *over*-sized negative margin (410px of pull-up against a 92px header). With an extra wrapper the pull-up is still correct and the clearance still fires; it just lands on the wrapper, so the wrapper's contents shift down together and the first block's background no longer runs under the header. The overlay degrades to looking non-overlaid, with content correctly placed. That's why I left it unguarded rather than heuristically descending to find the painted block — the mis-target is cosmetic, and a sniffing heuristic would be the fragile part.

**Low — `overlayTargetSelector` requires a direct child of `.wp-site-blocks`.** Leaving as-is, though your framing sharpened one detail worth recording.

The pull-up is scoped to `.wp-site-blocks > … + *`, so under a wrapped header it never applies and the variable has no consumer — measuring nothing is correct there. You're right that it silently no-ops, but the practical consequence is narrower than "the whole page": `position: fixed` on the header is *not* scoped that way, so a wrapped header still overlays; being out of flow, content starts at viewport top on its own and needs no pull-up. What it would want is the clearance, which does no-op.

That's a real limitation, but not a regression — before this PR the variable was set to whichever matching part came last (the footer), and the pull-up ignored it anyway, so a wrapped-header theme got no overlay clearance then either. Generalizing it means generalizing the CSS pull-up first; doing it in JS alone would reintroduce the second-writer bug this PR removes. Happy to pick that up as follow-up if a theme in that shape comes up.

All 19 checks green, including Security Audit after the `seroval` pin.

PR #504, opened 2026-08-12
chore(i18n): regenerate POT and complete JS translation catalog

### LINKED ISSUE #503: i18n: stale POT header, missing JS translation JSON files, and stray docs shipping in languages/
While auditing translation compliance I found three issues in `languages/`:

1. `designsetgo.pot` was generated from a `deployment-package/` path that doesn't exist in this repo (2129 bogus source-location comments) and its `Project-Id-Version` header is still `1.5.0` (plugin is at 2.6.3).
2. 6 newer script handles (`advanced-heading`, `background-video`, `text-style` format, `custom-css`, `heading-segment`, `clickable-group` — 54 files across 9 locales) never got their per-locale JS translation JSON files generated, so the catalog in `languages/` is incomplete relative to the current `.po` files.
3. `languages/FINAL-100-PERCENT-COMPLETE.md`, `TRANSLATION_STATUS_REPORT.md`, `TRANSLATION_SUMMARY.md`, and `TRANSLATION_README.md` are internal translation-progress working notes (stale stats, emoji progress reports) that aren't excluded by `.distignore`, so they currently ship inside release ZIPs.

None of these break existing translations — all currently-translated strings work correctly. This is a housekeeping/completeness fix.

miyanialkesh7:
## Description
Regenerates `languages/designsetgo.pot` from the actual plugin source (fixes bogus `deployment-package/` paths and the stale `1.5.0` version header), adds the 54 JS translation JSON files that were missing relative to the current `.po` files, and removes 4 internal progress-tracking docs that were shipping in `languages/` without being excluded from release ZIPs.

## Type of Change
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement

## Related Issue
Closes #503

## Changes Made
- Regenerated `languages/designsetgo.pot` from the real plugin source (correct paths, `Project-Id-Version: DesignSetGo 2.6.3`)
- Added 54 missing per-locale JS translation JSON files (6 script handles x 9 locales) so the catalog matches current `.po` content
- Removed `languages/FINAL-100-PERCENT-COMPLETE.md`, `TRANSLATION_STATUS_REPORT.md`, `TRANSLATION_SUMMARY.md`, `TRANSLATION_README.md`

## Testing
- [x] Ran `wp i18n make-pot`/`make-json` against a live local WP 7.0.3 instance and diffed output against committed files before changing anything
- [x] Verified all 9 locales' `.po`/`.mo` pairs still load correctly (untouched by this PR)
- [x] Activated the plugin on a live WP instance — no PHP errors/warnings in debug log
- [x] Ran `wp plugin check designsetgo` with the same exclusions as CI's `plugin-check.yml` — 0 new findings (one pre-existing, unrelated `phpstan-constants.php` warning only)
- [x] Validated JSON syntax of all 54 new files
- [ ] Tested in WordPress editor (no browser automation available in the environment this PR was prepared in — verified via WP-CLI/REST + debug log instead; a maintainer spot-check of a non-English editor locale is recommended)
- [x] Tested on frontend (site loads normally with plugin active, no errors)
- [x] No console errors
- [x] No PHP errors

## Screenshots/Videos
N/A — no UI changes, translation-catalog housekeeping only.

## Checklist
- [x] My code follows the WordPress Coding Standards (no PHP changed)
- [ ] I have commented my code, particularly in hard-to-understand areas (n/a — no code logic changed)
- [x] I have updated documentation as needed
- [x] I have tested on WordPress 6.4+ (tested on 7.0.3)
- [x] All files are under 300 lines (if applicable)
- [x] Accessibility: WCAG 2.1 AA compliant (n/a — no UI changes)
- [x] Security: All user input is validated and sanitized (n/a — no user input involved)
- [x] Internationalization: All user-facing strings use `__()` or `_e()` (this PR is the i18n catalog itself)

## Additional Notes
None of the affected strings were previously translated (they're newer blocks), so this PR doesn't change any user-visible translated text — it completes the generated-artifact set and cleans up the source `.pot`.


jnealey-godaddy:
Thanks for digging into this. The POT really was stale, and the four working-note docs shipping in the ZIP was a good catch.

I smoke tested the branch though, and the JSON part doesn't take effect. On a WP 7.0 instance switched to de_DE, `wp.i18n.getLocaleData('designsetgo')` in the editor reports 1006 messages on `main` and 1006 on this branch. None of the six target msgids show up on either.

It's the filename hash. `load_script_textdomain()` in `wp-includes/l10n.php` looks for `designsetgo-{locale}-{md5($relative)}.json`, and `$relative` is the *built* script path with `plugins/designsetgo/` stripped, so `build/blocks/advanced-heading/index.js`. The new files hash the source path:

```
md5('src/blocks/advanced-heading/edit.js')  = 791b7a2433658dc6eca454df19d848dc   <- generated here
md5('build/blocks/advanced-heading/index.js') = ba2faa51118a56c66d145208dc453c1e  <- what core opens
```

That traces back to the POT. The old one had 1909 `build/` references; this one has none, which means `make-pot` ran in a tree with no `build/` directory. `npm run i18n:make-json` takes its filenames from those references, so from that POT it can only ever emit source-keyed files. Running `npm run build` first sorts it out.

Worth knowing before you redo them: the `.po` files have `msgstr ""` for all six of these strings anyway, so even correctly keyed files wouldn't change what a German user sees. The blocker there is that nobody has translated them yet.

Could you drop the 54 JSON files from this PR? The POT regeneration and the doc deletions are both good, and I'd like to merge those.

Separately, poking at this surfaced something worse that isn't your doing. The German editor renders "Constrain Inner Width" as "Constrain Width" and "HTML Element" as "UI Elements". Those aren't translations, they're different English strings, so msgids and msgstrs are misaligned somewhere in the `.po` files. It's live on `main` right now. I'll open an issue.

miyanialkesh7:
Good catch on the hash mismatch — that makes sense given the POT was generated without a `build/` dir present. Dropped the 54 JSON files in 6f3b5086; the POT regeneration and doc deletions stay as-is.

Once the six strings actually have translations in the `.po` files, I'll regenerate the JSON catalog with `npm run build` first so the filenames hash against the built paths.

Thanks for flagging the `.po` misalignment on `main` too — that's outside this PR's scope, sounds like the right call to track it separately.

PR #514, opened 2026-08-16
feat: interaction layers extension

jnealey-godaddy:
Adds a global **Interaction Layers** extension: any block can bind a trigger (click, hover, scroll-into-view, exit intent, key press) to an action on a target it chooses, evaluated by one small delegated frontend runtime.

This generalises what we already had in one place — the animation extension's scroll trigger — into a vocabulary that works across all 69 blocks, and closes the largest capability gap against GreenShift.

## What an author gets

**Interactions** panel in block Settings, alongside Animations and Hover Effect. The sidebar holds a scannable list (`Click → Toggle class is-open`, with the target selector beneath); clicking a row opens a modal with the full configuration.

Targets can be picked visually — **Pick on canvas** shrinks the modal to a corner prompt, outlines blocks on hover, and writes the selector on click. Blocks with no anchor or class are given a generated `dsgo-target-xxxxxx` class so the selector is guaranteed to resolve rather than silently matching nothing. The raw selector field stays for advanced cases, with a live match count.

### Actions

| Group | Actions |
|---|---|
| Visibility | Show, Hide, Show / hide |
| Classes and attributes | Toggle / Add / Remove class, Set attribute, Remove attribute |
| Scrolling | Scroll to (with offset), Scroll to top of page |
| Blocks | Open modal, Close modal, Submit form |
| Media | Play, Pause, Play / pause video or audio |
| Other | Copy to clipboard, Move focus to, Fire a custom event |

## Design decisions worth reviewing

**One `dsgoInteractions` array attribute, not flat attributes per action.** The animation extension uses eight flat attributes; that shape does not extend, because every new action would need a new attribute on every block and every rename a deprecation across all 69. A single object-typed attribute keeps the schema fixed as the vocabulary grows.

**No deprecation needed.** `data-dsgo-interactions` is omitted from `save()` entirely when the list is empty, so existing content serialises byte-identically. The 454 existing deprecation round-trip tests pass unchanged, and a reloaded editor shows no "Attempt Recovery".

**Conditional loading.** The runtime is 3 KB and only enqueued when rendered block content actually contains `data-dsgo-interactions`. Pinned by an e2e test.

**Accessibility is handled by the runtime, not left to the author.** A click-triggered element that is not already a button or link gets `tabindex="0"` and `role="button"`, and Enter/Space activate it. Visibility actions mirror the resulting state onto the trigger as `aria-expanded` — but only on elements that present as controls, since the attribute is noise on a plain div.

**One delegated listener per event type on `document`.** Not one listener per element behind a shared flag — that pattern starves every element after the first and breaks silently after a DOM swap, which has bitten us before in the sticky-header work.

## Changes outside the extension

- `src/blocks/modal/view.js` — the modal *dispatched* `dsgo-modal-open` / `dsgo-modal-close` but never listened for them, so the open/close actions had nothing to talk to. Added document-level listeners. A modal's own announcement bubbles to `document` and would re-enter the handler, so commands are distinguished from announcements by `e.target === document`. Verified against a real modal: programmatic control works, and the normal trigger path still fires exactly one event.
- `src/extensions/visibility/filters.js` — rendered its controls bare into `group="advanced"` with no `PanelBody`, so "Always visible" and "Add rule" read as part of "Additional CSS class(es)". Now titled **Visibility**, matching Style Bindings.
- `includes/core/class-assets.php` — registers and conditionally enqueues the runtime.
- `webpack.config.js` — one new entry.

## Testing

- **73 new unit tests** across constants, target resolution, the action registry, attribute/save-prop registration, the panel, the summariser, the canvas picker, and the frontend runtime. Full suite: **10,335 passing**.
- **6 new PHPUnit tests** for the server-side attribute mirror. Full suite: 1,004 passing.
- **3 new e2e specs** driving editor → publish → frontend, including a check that the runtime is *not* requested on a page without interactions.
- Manually verified in the editor and on the frontend: panel placement, modal, canvas picker, hide/show with `aria-expanded`, keyboard operation, and a clean console.

Three of the tests were written after finding real bugs, and each was confirmed to fail when the fix is reverted:

- The canvas picker let the editor select the block on `mousedown`, which re-rendered the inspector for the newly selected block and destroyed the modal mid-edit. The runtime now swallows the whole pointer sequence while picking.
- `actions.js` imported a constant from a module that imports `@wordpress/i18n`; webpack externalises that to `wp.i18n`, which the dependency-free frontend bundle never loads — every interaction on the site died with a console error. The constant moved to a leaf module, script registration now reads the generated `.asset.php`, and a test walks the frontend module graph rejecting any `@wordpress/` import.
- `attributeName` was overloaded for both the keyboard key and the attribute name. Split into a separate `key` field before any content could store the ambiguous shape.

## Follow-ups, deliberately not in this PR

- **Tabs and accordion control.** Tabs dispatches `dsgo-tab-change` but doesn't listen for it; accordion has no event API. Both need the same listener work the modal just got.
- **Conditions** between trigger and action (`if viewport < X`, `if logged in`). The attribute shape has room for a `conditions` key.
- **`inView` offset.** The offset field currently applies only to `scrollTo`; the in-view observer uses a fixed `rootMargin`.
- **Persistent dismiss** (hide + remember). Better as a `persist` flag than a separate action.


jnealey-godaddy:
Thanks — all seven were real. Verified each against the code before fixing, and the first one was the serious find.

## Blocking

**Dynamic blocks (save-props.js:23)** — confirmed. `save()` returns `null` on all 28 blocks this plugin renders server-side, so `getSaveContent.extraProps` never runs and the attribute never reached the frontend. New `DesignSetGo\Interactions` (`includes/features/class-interactions.php`) injects it at `render_block` priority 5 via `WP_HTML_Tag_Processor` — ahead of the priority-10 enqueue check that decides whether to load the runtime. It skips markup that already carries the attribute, so static blocks are untouched, and it reduces stored attrs to the known schema rather than echoing arbitrary JSON back out. 9 PHPUnit tests. Verified live: a `designsetgo/breadcrumbs` block (stored as a bare self-closing comment, no save output at all) now renders `<nav data-dsgo-interactions=…>` and toggles its target.

**Unvalidated class/attribute values (actions.js:82)** — confirmed. Added `parseClassNames()`, which strips a leading dot and splits on whitespace, so `.is-open` and `a b` both work instead of throwing. `isAttributeAllowed()` now also rejects names that aren't valid XML names — `setAttribute('1 bad', …)` was the other throw path. Belt and braces: `fire()` wraps each `runAction` in try/catch, since anything that escapes there aborts every interaction queued behind it in the same delegated listener.

**Hover re-fires (frontend.js:132)** — confirmed. Capture delivers `mouseenter` for descendants, so `closest()` resolved back to the same source on every child crossing. Now requires `e.target` to be the element carrying the attribute. Verified in a browser with two child spans: the toggle stays on.

## Correctness

**keydown unreachable (frontend.js:212)** — confirmed, and I went further than adding `tabindex`. Scoping key presses to the focused element is the wrong model: "press Escape to close" should work wherever focus is. `keydown` is now page-level. Enter/Space activating a synthesised button stays focus-scoped, since that stands in for a real click.

**submitForm (actions.js:115)** and **media (actions.js:130)** — confirmed, both now look down as well as up. `core/video` renders `<figure><video>`, so the block an author targets is never the media element itself.

**derive-selector.js:68** — confirmed. Now checks `hasBlockSupport(name, 'customClassName', true)` and returns no selector rather than a dead one, with a warning Notice in the modal telling the author to use an HTML anchor instead.

## Convention notes

**PanelBody** — checked this rather than assuming: `DsgoInspectorPanel` is used by 83 block files and **zero** extensions. Animations, Hover Effect, Custom CSS and Style Bindings all use `PanelBody`. That wrapper enforces the per-block Settings/Style/Advanced IA, which doesn't apply to a cross-cutting extension panel. Keeping `PanelBody` — and the `visibility/filters.js` change now matches `style-binding` exactly.

**Initial hidden state** — you're right that there's no UI for it. Added an info Notice naming the exact class when the action is `show`. A real "start hidden" control belongs on the target block, not the trigger, so I've left it as a follow-up rather than bolt it on here.

## Also found while in there

- **`setAttribute` allowed `javascript:` URLs.** The `on*` blocklist didn't cover `href`/`src`/`formaction`/etc., so `setAttribute('href', 'javascript:…')` was a live XSS vector. Now rejects `javascript:`, `data:` and `vbscript:` on URL-bearing attributes; ordinary URLs still pass.
- **`inView` + `once` never unobserved**, costing a callback on every scroll past the element for the life of the page.
- **`aria-expanded` was absent until first click.** Visibility controls now publish their starting state on init, so a screen-reader user isn't told about a collapsed region only after operating it.
- Moved `HIDDEN_CLASS`/`VISIBILITY_ACTIONS` into `visibility-contract.js`. `constants.js` imports `@wordpress/i18n`, which webpack externalises to `wp.i18n` — a global the dependency-free frontend bundle never loads. One import of it from the runtime throws on load and disables every interaction site-wide. There's a test that walks the frontend module graph and rejects any `@wordpress/` import.

**Gate:** build · lint:js/css · phpcs · **10,355 unit tests** · **1,013 PHPUnit** · 3 e2e — all green. Browser-verified the dynamic-block injection and the hover fix.

The one remaining red check, Plugin Check, is a pre-existing false positive on `includes/blocks/forms/class-form-security.php:147` — the Turnstile server-side verification call, byte-identical on `main` and untouched by this PR.

jnealey-godaddy:
All five confirmed and fixed in `67f80e44`.

## High — `srcdoc`

Correct, and the framing was the useful part: the existing checks covered two of the three ways an attribute can introduce script, and `srcdoc` is the third. Rather than patch the one name, the guard now names all three classes explicitly so the next addition has an obvious home:

1. **Event handlers** — the value *is* script (`on*`).
2. **Raw-HTML attributes** — the value is parsed as markup, so no scheme check can help (`srcdoc`).
3. **URL attributes** — the value is fetched or navigated to, so a `javascript:` scheme executes.

Also added an element-level guard while in there. `setAttribute` now refuses to write to `<script>`, `<base>` or `<meta>` at all: `<base href>` silently repoints every relative URL on the page and `<meta http-equiv="refresh">` redirects it. Neither is a legitimate interaction target, and both are reachable through attribute names that pass all three checks above.

Verified in a browser against the real built bundle, configured exactly as described (`targetSelector: '#victim'`, `attributeName: 'srcdoc'`): the attribute is rejected, no script executes, and a benign `title` interaction in the same list still applies — so the guard rejects the interaction, not the element.

## Medium — `sanitize()` rationale

Fair, and the docblock was the actual defect. It claimed a security rationale that the early return makes false for static blocks. Rewritten to say what it really is: a normalisation step for the dynamic path only, giving that path a fixed shape and known types — explicitly *not* a filter over stored content generally. Added a pointer at the early return so the two halves stay honest about each other.

I kept the early return rather than sanitising static markup too. Re-encoding stored attributes at render time would mean the markup in the database and the markup served could drift, which is a worse problem than the one it would solve.

## Medium — non-scalar cast

Confirmed. Extracted `string_field()`, which returns the default for anything non-scalar instead of casting, and `offset` now requires `is_numeric` before the float cast. Two PHPUnit tests cover it, including an array reaching `id` and `targetSelector`.

## Low — clipboard, keydown scan

Both fixed. `writeText()` now has a `.catch()`. `keydownTargets()` became a cached list rebuilt in `initInteractions()`, so a keystroke iterates a prepared array instead of querying the DOM and re-parsing every interaction's JSON — that ran on every keypress including while typing in a form field. Since caching risks going stale, there's a test that a keydown interaction added by a soft-reload DOM swap still fires.

**Gate:** build · lint:js/css · phpcs · **10,363 unit tests** · **1,015 PHPUnit** · 3 e2e — all green.

jnealey-godaddy:
All three confirmed and fixed in `3bbc3617`.

## High — control-character evasion

Confirmed the regex miss, then confirmed it's genuinely exploitable rather than just a theoretical gap. Chrome resolves both variants back to a script URL:

```js
const a = document.createElement( 'a' );
a.setAttribute( 'href', 'jav' + String.fromCharCode( 9 ) + 'ascript:alert(1)' );
a.protocol; // => "javascript:"
```

Fixed by normalising the value the way the URL parser does before testing the scheme, rather than by extending the pattern. Extracted as `isDangerousUrl()`, which removes every ASCII control character and space
(code points 0x00–0x20, plus 0x7F) from the value, then tests the remainder
against `/^(javascript|data|vbscript):/i`. The scheme is decided on the
normalised copy; the attribute is still written with the original value.

Two notes on the shape of that:

- It strips **every** C0 control and space, not just tab/LF/CR. The spec only removes those three from the middle, but being stricter than the parser can only ever reject a URL whose *scheme* contains a control character — which no legitimate URL has. That avoids this needing another patch if a parser difference turns up.
- Stripping is used only to decide the scheme. The value itself is still written unchanged, so `https://example.com/a b` is unaffected.

25 unit tests, including the tab/LF/CR splits, a split before the colon, multiple splits in one value, a leading NUL, a split `data:` URL, and negative cases (`mailto:`, `tel:`, and URLs merely *containing* "javascript:" in a path or query). Confirmed the suite catches the regression by restoring the old regex — 9 of the 25 fail.

Also verified against the real built bundle in Chrome, configured as an author would: both evasions rejected, and a legitimate `https://` value in the same interaction list still applied.

## Low — observer leak

Confirmed. `IntersectionObserver` holds strong references to its targets, so every element replaced by a soft-reload navigation stayed alive with its subtree for the life of the tab. The runtime now tracks observed elements and unobserves any that are no longer `isConnected` on each `initInteractions()` — which already runs on `dsgo-content-loaded`.

Precise rather than a blanket `disconnect()`, since `initInteractions()` accepts a subtree root and disconnecting would silently drop elements outside it. New test file covers observe/unobserve across a swap, survivors not being dropped, one-shot unobserving after firing, repeatable interactions staying observed, and the no-IntersectionObserver path. It's a separate file because the observer lives in module scope and needs a fresh registry.

## Low — JSDoc

Fixed — the `@return` tag had a duplicated `|null}` fragment, damage from an earlier scripted edit.

**Gate:** build · lint:js/css · phpcs · **10,395 unit tests** · **1,015 PHPUnit** · 3 e2e — all green.

PR #516, opened 2026-08-17
feat: off-canvas panel mode for the modal block

jnealey-godaddy:
Adds an edge-anchored slide-in panel mode to `designsetgo/modal`, exposed as an **Off-Canvas Panel** block variation. Reuses the modal's existing focus trap, scroll lock, Escape handling, and trigger machinery — no new frontend JS, no new dependencies.

Implements plan 5 of the GreenShift gap roadmap.

## What's new

Three attributes on the modal — `displayMode` (`dialog` | `panel`), `panelEdge` (`left` | `right` | `top` | `bottom`), and `panelSize` (any CSS length) — plus the variation, the inspector controls, and the CSS that positions and animates the panel.

In panel mode the inspector swaps the dialog sizing controls (Width / Max Width / Height / Max Height, all inert for a panel) for Slides In From and Panel Size, and swaps them back on return to dialog mode.

## No deprecation required — and that's verified, not assumed

All three attributes have defaults and every existing modal leaves them there, so WordPress omits them from the block comment. `save()` emits the panel class and the `--dsgo-panel-size` property **only** when `displayMode` differs from its default, so stored markup for an existing modal is character-identical.

The 820-test deprecation guard suite (`deprecations-isEligible`, `modal`, `modal-overlay-deprecation`, `block-attributes`) passes untouched, and `tests/unit/modal-panel-save.test.js` pins the invariant directly: a default modal serialises with no panel class, no panel attributes, and no custom property.

## Three departures from the plan

The plan was written against an assumed DOM. Each of these was found while implementing:

1. **The mode classes go on the block root, but every positioning rule targets `.dsgo-modal__dialog`.** `.dsgo-modal` is the full-screen fixed overlay that also contains the backdrop; the box that actually slides is the dialog. Styling the root would have fought the overlay.

2. **The variation ships `animationType: 'fade'`, not `'none'`.** `dsgo-modal--animation-none` sets `transition: none !important` on `.dsgo-modal__dialog` — the same element the panel transform animates — so `'none'` would have silently suppressed the slide. `'fade'` animates opacity only, so the two compose. The panel's state rules carry the edge class (specificity 0,4,0) so they still beat the `slide-*` / `zoom` transforms if an author picks one.

3. **`closeButtonPosition` is `inside-top-right`.** The plan's `top-right` sits at `-12px`, which a screen-edge panel clips.

## A bug caught in the browser, not by the tests

The first cut declared `--dsgo-panel-size: 24rem` on `.dsgo-modal__dialog` as a default. `save()` writes that property on the **root**, so the declaration shadowed the inherited value and **every panel rendered at 24rem regardless of what the author set**.

It is invisible at the default size — the unit tests and the first browser pass both used it. It only surfaced when deliberately testing an oversized panel. The default now lives in the `var()` fallback at each use site, and a unit test pins the invariant; that test was confirmed to fail when the declaration is reintroduced.

## Verification

Browser-verified against the built assets on wp-env, not just in jsdom:

- All four edges land flush at the correct dimensions (left/right full viewport height, top/bottom full width).
- Focus stays inside the panel; body scroll locks and unlocks.
- Escape and backdrop click both close.
- `prefers-reduced-motion` yields no transform and no transition — the panel appears in place.
- An oversized panel clamps to the viewport with no horizontal page scroll.
- Author-set `panelSize` is honoured (20rem → 320px), default still applies when unset.
- **Regression:** an ordinary centred modal is unchanged — still `position: relative`, still centred, no panel class, no validation warning.

`openModal` from the Interaction Layers extension drives the panel with no extra wiring, so a hamburger button is genuinely a one-liner: a click interaction with action `openModal` and the panel's modal ID as the value. `tests/e2e/offcanvas-panel.spec.js` covers that path.

## Tests

- `tests/unit/modal-panel-save.test.js` — 22 tests: default-mode markup identity, panel class/property emission, four-edge round-trip validity, unrecognised-edge clamping, dialog semantics preserved, no inline dimensions in panel mode, the property landing on the root rather than being relocated onto `.dsgo-modal__content` by `transferStylesToContent()`, and the custom-property shadowing guard.
- `tests/phpunit/block-inserter-modal-panel-test.php` — 10 tests pinning the Abilities block-inserter output against `save()`: mode classes per edge, the size property's position relative to the class attribute, edge clamping, and the absence of inline content dimensions.
- `tests/e2e/offcanvas-panel.spec.js` — 4 specs: edge geometry + focus trap + scroll lock, content fill on a full-width bottom panel, open via an `openModal` interaction, and the centred-modal regression.

Gates: `build`, `lint:js`, `lint:css`, `phpcs` (121 files), **13,245 unit tests**, **1,028 PHPUnit**, and the e2e specs all pass.

## Review rounds

Seven review findings, all confirmed in a browser before fixing.

**Panel content did not fill the panel (High).** `save()` wrote the dialog-mode `width:600px;max-width:90vw` inline onto `.dsgo-modal__content`, which outranks the stylesheet rule. The content is always 648px and only *appears* correct while the panel is narrower and flex-shrink absorbs it — so the default right panel looked fine and a bottom panel (which spans the viewport) rendered 648px of content pinned left. Panel mode now passes no dimensions to `transferStylesToContent()`.

**Content overflowed the dialog by its padding.** Surfaced by the fix above and not part of the review: `.dsgo-modal__content` is content-box, so `height: 100%` plus 24px padding overflowed by 48px and pushed content past the viewport edge. Added `box-sizing: border-box`.

**The Abilities block-inserter had drifted (High).** `includes/abilities/class-block-inserter.php` hand-builds modal HTML that must byte-match `save()`. It emitted no panel classes, no `--dsgo-panel-size`, and kept the inline dimensions, so a panel-mode modal inserted that way would fail block validation on first edit. Now mirrored, including that the style attribute follows the class attribute.

**An unrecognised `panelEdge` produced a broken panel (Low).** Reachable via `createBlock()`, hand-edited content or third-party code, though not the inspector. Confirmed as a 207x123 box floating mid-viewport with no validation error. `block.json` now declares enums, and — since an enum is advisory and cannot stop `createBlock()` — `save()` clamps to a known edge, with the PHP mirroring it.

**A `max-width: 480px` block was a no-op (Low).** `max-width`/`max-height: 100%` on the dialog already clamp an oversized panel at every viewport width. Removed, after reproducing the clamp with it deleted.

Each fix has a test that was confirmed to fail when the fix is reverted.

## On the variation's scope

The nine existing modal variations use `scope: ['block']` — content templates offered inside the modal's own placeholder once you have already chosen a modal. This one uses `scope: ['inserter', 'transform']` deliberately: an off-canvas panel is a different structural choice, and an author wanting a slide-in menu searches the inserter for "panel", not "modal".

## Out of scope

- Swipe-to-dismiss on touch.
- A push variant where page content shifts rather than the panel overlaying it.
- An editor canvas preview reflecting panel mode. The inspector controls swap correctly, but the canvas still renders the centred `dsgo-modal-editor-preview` chrome in both modes.

## Note on local e2e

The e2e suite was failing on this machine before any of this work — modal *and* form-builder alike. `.wp-env.override.json` points at a WP 7 core checkout and the DB had a nonexistent `default` theme active, so core's block-asset URL resolution fell through to the theme branch and enqueued every DSGo asset at a broken URL; zero of 65 blocks registered in the editor. `wp theme activate twentytwentyfive` fixes it. That is local DB state, not a repo change, so it will recur if that install is reset.


jnealey-godaddy:
All seven points confirmed and addressed. Two Highs in `7cfcf780`, the Lows in `a4280f45`.

## High — inline width/maxWidth beats the panel's `width: 100%`

Confirmed, and it is worse than "the content renders at 600px". I measured the built page in a browser before changing anything:

| panel | dialog | content | fills? |
|---|---|---|---|
| right 24rem | 384 | 384 | yes |
| right 40rem | 640 | 640 | yes |
| **bottom, 1400px viewport** | **1400** | **648** | **no — pinned left** |

The content is always 648px (600 + 48 padding) and only *appears* correct while the panel is narrower and `flex-shrink` absorbs it. So the default right-hand panel from the shipped variation looks fine, and the bug only bites once the panel is wider than 648px — which a top or bottom panel always is, since it spans the viewport. That is why my original browser pass missed it: I measured `.dsgo-modal__dialog` geometry and never the content's.

Fixed exactly as suggested — `transferStylesToContent(blockProps, isPanel ? {} : { width, maxWidth, height, maxHeight })` — with `edit.js` mirrored so the editor preview matches.

**Fixing that surfaced a second bug the review did not mention.** With the inline width gone, `.dsgo-modal__content` is content-box, so `height: 100%` plus the 24px padding overflowed the dialog by 48px and pushed content past the viewport edge (922px of content in an 874px dialog). Width had been surviving that only because flex-shrink clamps the main axis. Added `box-sizing: border-box`; both axes now match the dialog exactly.

The e2e test you noted as missing now exists and asserts on `.dsgo-modal__content`, on a bottom panel specifically. I verified it catches the regression by reverting the fix — it reports `widthDelta: 680`.

## High — the same issue seen from style.scss

Same root cause, same fix. Nothing separate to change.

## Low — no `enum` on `displayMode` / `panelEdge`

Confirmed, and I checked what actually happens rather than assuming. Forcing `dsgo-modal--panel-diagonal` onto a real panel produced `position: fixed` with no offsets resolved — a **207x123 box floating mid-viewport**, no validation error. Broken, though not invisible.

Added the enums as suggested. But an enum in `block.json` is advisory — it does not stop `createBlock()` writing a bad value — so `save()` now also clamps to a known edge, keeping the broken markup unreachable rather than merely undeclared. `class-block-inserter.php` mirrors the clamp. Covered by unit tests (`diagonal`, `''`, `RIGHT`, `left; }`, `undefined`) and a PHPUnit data provider.

One note on the test: asserting `not.toContain('dsgo-modal--panel-' + edge)` passes vacuously for `edge === ''`, because that string is a substring of the correct `dsgo-modal--panel-right`. The test matches the emitted class instead.

## Low — redundant `max-width: 480px` block

Confirmed and removed. `max-width: 100%` / `max-height: 100%` on the dialog are unconditional, so an oversized panel is clamped at every viewport width. Verified empirically both before and after: a 60rem panel clamps to the viewport at 1400px wide (nowhere near the breakpoint), and still clamps with the block deleted. The content rules it repeated (`max-height: 100%`, `border-radius: 0`) were already unconditional too.

## Low — dangling plan-doc reference

Correct, dropped. The plan docs are deliberately not part of this PR.

## Low — variation scope deviates from the other nine

Deliberate, and now documented in a comment at the call site plus the PR description. The nine existing variations are content templates offered inside the modal's own placeholder once you have already chosen a modal. An off-canvas panel is a different structural choice — an author wanting a slide-in menu searches the inserter for "panel", not "modal" — so it is exposed as its own inserter entry and transform target.

---

**Gate:** build, `lint:js`, `lint:css`, `phpcs` (121 files), **13,245 unit tests**, **1,028 PHPUnit**, 8 off-canvas e2e specs — all green.

jnealey-godaddy:
Both confirmed and fixed in `97f094c5`.

## Low — `panelSize` CSS injection

Confirmed, and reproduced before changing anything. A `panelSize` of `10px;position:fixed;inset:0;background:red` emits:

```
style="--dsgo-panel-size:10px;position:fixed;inset:0;background:red"
```

Four declarations on the modal root, not one. React does not escape `;` inside a style value, and `esc_attr()` on the PHP side stops an attribute break-out but not a declaration break-out — so your read of the mechanism is exactly right.

Agreed on the severity too: it needs post-edit capability and yields CSS rather than script. It is not nothing, though — `position: fixed; inset: 0` on the modal root is a clickjacking primitive, and a `url()` value leaks a request on render.

Fixed in both `save.js` and `class-block-inserter.php`: a single plain CSS length is allow-listed, anything else falls back to `24rem`.

**Deliberately an allow-list, not a denylist.** A denylist of dangerous substrings is precisely what had to be patched three times in a row on the interaction-layers PR (`srcdoc`, then control-character scheme evasion), so a pattern that enumerates what is *permitted* is the durable shape here.

One trade-off worth stating: this also rejects `calc()`, `min()` and `clamp()`. The Panel Size `UnitControl` cannot produce those, so nothing in the UI regresses — but a programmatic caller passing `calc(100% - 2rem)` now silently gets `24rem` instead. I judged that the right default given the attribute exists to hold a control's output; happy to widen it to a parsed `calc()` if you would rather keep the expressiveness.

Covered by 14 unit tests (6 valid lengths pass through unchanged, 8 crafted values fall back) and 13 PHPUnit tests over the same matrix, which additionally assert the injected declarations are absent from the output.

You are also right that this mirrors the pre-existing lack of validation on `width`/`maxWidth`/`height`/`maxHeight`. I have deliberately left those alone — they are on the dialog-mode path this PR does not touch, and tightening them is a behaviour change for existing published content that deserves its own PR rather than riding along here.

## Low — duplicate `prefers-reduced-motion` rule

Confirmed dead and removed. Verified the selector rather than reasoning about it: a panel dialog returns `true` for `.matches('.dsgo-modal .dsgo-modal__dialog')`, because the panel root carries `dsgo-modal` and `dsgo-modal--panel` together — so the block-wide rule at the top of the file already applied.

Re-verified the behaviour after deleting it, in a reduced-motion browser context: `transform: none`, `transition-duration: 0s`, panel flush to the right edge (`left: 896` at a 1280px viewport, width 384). It appears in place with no slide, as before.

Left a comment in its place explaining why no panel-specific rule is needed, so it does not get re-added.

---

**Gate:** build, `lint:js`, `lint:css`, `phpcs` (121 files), **13,259 unit tests**, **1,041 PHPUnit**, 8 off-canvas e2e specs — all green.

PR #518, opened 2026-08-17
feat: schema.org JSON-LD output for the accordion (FAQ, How-to)

jnealey-godaddy:
Lets an author opt a block into emitting schema.org JSON-LD. One `<script type="application/ld+json">` per page, nothing emitted unless an author deliberately chooses a type.

Implements plan 4 of the GreenShift gap roadmap.

## What's new

A `dsgoSchema` attribute on `designsetgo/accordion`, offering **FAQ** (`FAQPage`) or **How-to** (`HowTo`), with a Structured Data control in the block's Advanced panel alongside Visibility and Style Bindings. `DesignSetGo\SchemaOutput` walks the singular post's blocks on `wp_head` and prints a single merged `@graph`.

## Two plan assumptions that were wrong

Both were verified against the real blocks before writing any code.

**1. The accordion title is HTML-sourced — the planned builder would have emitted nothing.**

The plan read the question from `$block['attrs']['title']`. But `accordion-item` declares:

```json
"title": { "type": "string", "source": "html", "selector": ".dsgo-accordion-item__title" }
```

`parse_blocks()` only decodes the block comment's JSON, so that key is **absent for every real post**. The builder would have found an empty question on every item, skipped them all, and silently emitted nothing — while passing any test that hand-builds the array with `attrs.title` set, which is exactly what the plan's tests did.

The title is now read from the saved markup with `WP_HTML_Processor` (chosen over `DOMDocument`, which mangles UTF-8 without an encoding preamble, and over a regex, which cannot track nesting). Every builder test drives `parse_blocks()` output rather than invented attributes, and one test asserts `attrs.title` is absent so the regression cannot return unnoticed.

**2. Review/Product has no data source, so it is not in this PR.**

The plan mapped Review onto `comparison-table` and `card`. The real columns are `{name, link, linkText, featured}` — no `title`, and **no rating field anywhere in either block**. `product-showcase-hero` has only a `showRating` display toggle over WooCommerce data, which WooCommerce already emits schema for.

A control offering a type with no builder behind it is a no-op that looks like a feature, so `SCHEMA_TYPES` lists only the accordion, and a test pins the allowlist to blocks that actually have a builder. Review can be its own PR once a rating source exists.

## Why opt-in, and why one script

Emitting FAQ markup for an accordion that holds navigation links is structured-data spam and gets sites penalised, so the default is `none`. Google also caps FAQ rich results to one block per page, so the collector merges every opted-in block into a single `@graph` rather than printing several scripts.

The control's help text tells authors to leave it on None if an SEO plugin already outputs the type for that page.

## Design notes

- **Reads stored content, not `render_block`.** The accordion is static (a `save.js`, no `render.php`) and its title only exists in saved markup; `wp_head` also runs before the content, so a `render_block` collector would be too late.
- **Cheap bail before parsing.** `strpos( $content, 'dsgoSchema' )` short-circuits before `parse_blocks()` on the overwhelming majority of posts.
- **Script break-out is neutralised.** `JSON_UNESCAPED_SLASHES` keeps URLs readable but means a literal `</script>` in a question would close the element early, so slashes are escaped after encoding. A test feeds `</script><script>alert(1)</script>` through a question title and asserts the output still parses.
- **`designsetgo_schema_nodes`** filter can modify or suppress the graph.
- Text is separated at tag boundaries before stripping — without it two adjacent paragraphs fused into `One.Two.`

## Verification

Browser-verified end to end on wp-env — real editor → save → `parse_blocks()` → `wp_head`:

```json
{"@context":"https://schema.org","@graph":[
  {"@type":"FAQPage","mainEntity":[
    {"@type":"Question","name":"What is DesignSetGo?",
     "acceptedAnswer":{"@type":"Answer","text":"A WordPress block plugin."}},
    {"@type":"Question","name":"Does it need jQuery?", ...}]},
  {"@type":"HowTo","name":"How to install DesignSetGo","step":[
    {"@type":"HowToStep","position":1,"name":"Download", ...}]}]}
```

Confirmed in that run: exactly **one** script tag; inline `<strong>` stripped from the question; `HowTo.name` taken from the post title (Google requires it); numbered steps; a Unicode arrow preserved; and a third, opted-out accordion on the same page contributing nothing. The inspector control was also driven like a user would — changing the select updates the attribute and back again.

## Tests

- `tests/phpunit/schema-builders-test.php` — 17 tests, all driven from `parse_blocks()` output: FAQ/HowTo shape, the HTML-sourced-title regression guard, inline markup stripped, entities decoded, the icon span excluded, incomplete pairs skipped while valid siblings survive, whole-class-name matching, and malformed HTML tolerated.
- `tests/phpunit/schema-output-test.php` — 14 tests: opt-in/opt-out, one script for two blocks, nesting inside a section, HowTo vs FAQ, `HowTo.name` from the title, valid JSON, script break-out, archives, empty blocks, and the filter modifying and suppressing output.
- `tests/unit/extensions/schema-attribute.test.js` — 9 tests: allowlist membership and exclusions, option list, `none` default, existing attributes preserved, and the allowlist pinned to blocks with builders.

Gates: `build`, `lint:js`, `phpcs` (124 files), **13,268 unit tests**, **1,072 PHPUnit** — all green.

## Not done

- Validation against <https://validator.schema.org/>. That means sending page content to a third-party service, so I left it as a manual step rather than doing it unprompted. The output matches Google's documented FAQPage and HowTo shapes.
- `Event` and `Recipe` types.
- Review/Product, per above.

### DIFF
--- a/includes/class-plugin.php
+++ b/includes/class-plugin.php
@@ -499,6 +499,13 @@ class Plugin {
 	public $interactions;
 
 	/**
+	 * Schema JSON-LD output instance.
+	 *
+	 * @var SchemaOutput
+	 */
+	public $schema_output;
+
+	/**
 	 * Block Bindings Support instance.
 	 *
 	 * @var Block_Bindings_Support
@@ -658,6 +665,7 @@ class Plugin {
 		require_once DESIGNSETGO_PATH . 'includes/features/class-extension-attributes.php';
 		require_once DESIGNSETGO_PATH . 'includes/features/class-style-binding.php';
 		require_once DESIGNSETGO_PATH . 'includes/features/class-interactions.php';
+		require_once DESIGNSETGO_PATH . 'includes/features/class-schema-output.php';
 		require_once DESIGNSETGO_PATH . 'includes/bindings/class-block-bindings-support.php';
 		require_once DESIGNSETGO_PATH . 'includes/data/svg-pattern-data.php';
 		require_once DESIGNSETGO_PATH . 'includes/features/class-svg-pattern-renderer.php';
@@ -710,6 +718,7 @@ class Plugin {
 		$this->extension_attrs        = new Extension_Attributes();
 		$this->style_binding          = new StyleBinding();
 		$this->interactions           = new Interactions();
+		$this->schema_output          = new SchemaOutput();
 		$this->block_bindings_support = new Block_Bindings_Support();
 		$this->block_bindings_support->register();
 		$this->modal_hooks      = new Blocks\Modal_Hooks();
--- a/includes/extension-configs/schema.php
+++ b/includes/extension-configs/schema.php
@@ -1,0 +1,25 @@
+<?php
+/**
+ * Schema extension attribute schema.
+ *
+ * Mirrors the JS allowlist so the attribute survives server-side block
+ * registration. Keep this list identical to SCHEMA_BLOCKS.
+ *
+ * @see src/extensions/schema/constants.js
+ * @package DesignSetGo
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+return array(
+	'blocks'     => array(
+		'designsetgo/accordion',
+	),
+	'exclude'    => array(), // phpcs:ignore WordPressVIPMinimum.Performance.WPQueryParams.PostNotIn_exclude -- Extension block-exclusion list, not a get_posts() query.
+	'attributes' => array(
+		'dsgoSchema' => array(
+			'type'    => 'string',
+			'default' => 'none',
+		),
+	),
+);
--- a/includes/features/class-schema-output.php
+++ b/includes/features/class-schema-output.php
@@ -45,13 +45,23 @@ class SchemaOutput {
 	 *
 	 * @param array  $blocks Parsed blocks.
 	 * @param string $title  Page title, passed to builders that need it.
+	 * @param array  $seen   Synced-pattern IDs already expanded, guarding cycles.
 	 * @return array List of schema graph nodes.
 	 */
-	public function collect( array $blocks, $title = '' ) {
+	public function collect( array $blocks, $title = '', array $seen = array() ) {
 		$nodes = array();
 
 		foreach ( $blocks as $block ) {
 			if ( ! is_array( $block ) ) {
+				continue;
+			}
+
+			// A synced pattern keeps its markup in a wp_block post; the page
+			// only stores a reference. Without expanding it, an FAQ defined
+			// once in a shared pattern would emit nothing on every page that
+			// uses it, with nothing to tell the author why.
+			if ( 'core/block' === ( isset( $block['blockName'] ) ? $block['blockName'] : '' ) ) {
+				$nodes = array_merge( $nodes, $this->collect_from_reference( $block, $title, $seen ) );
 				continue;
 			}
 
@@ -72,11 +82,48 @@ class SchemaOutput {
 			}
 
 			if ( ! empty( $block['innerBlocks'] ) && is_array( $block['innerBlocks'] ) ) {
-				$nodes = array_merge( $nodes, $this->collect( $block['innerBlocks'], $title ) );
+				$nodes = array_merge( $nodes, $this->collect( $block['innerBlocks'], $title, $seen ) );
 			}
 		}
 
 		return $nodes;
+	}
+
+	/**
+	 * Expand a synced pattern reference and collect from its content.
+	 *
+	 * @param array  $block Parsed core/block block.
+	 * @param string $title Page title.
+	 * @param array  $seen  Reference IDs already expanded on this branch.
+	 * @return array List of schema graph nodes.
+	 */
+	private function collect_from_reference( array $block, $title, array $seen ) {
+		$ref = isset( $block['attrs']['ref'] ) ? (int) $block['attrs']['ref'] : 0;
+
+		// A pattern that references itself, directly or through another
+		// pattern, would otherwise recurse until the request dies.
+		if ( $ref <= 0 || in_array( $ref, $seen, true ) ) {
+			return array();
+		}
+
+		$seen[] = $ref;
+
+		$reference = get_post( $ref );
+
+		if ( ! $reference instanceof \WP_Post || 'wp_block' !== $reference->post_type ) {
+			return array();
+		}
+
+		if ( 'publish' !== $reference->post_status ) {
+			return array();
+		}
+
+		if ( false === strpos( $reference->post_content, 'dsgoSchema' )
+			&& false === strpos( $reference->post_content, 'wp:block ' ) ) {
+			return array();
+		}
+
+		return $this->collect( parse_blocks( $reference->post_content ), $title, $seen );
 	}
 
 	/**
@@ -93,9 +140,11 @@ class SchemaOutput {
 			return;
 		}
 
-		// Cheap bail before the comparatively expensive parse. Every opted-in
-		// block carries the attribute name in its block comment.
-		if ( false === strpos( $post->post_content, 'dsgoSchema' ) ) {
+		// Cheap bail before the comparatively expensive parse. An opted-in block
+		// carries the attribute name in its block comment — unless it lives in
+		// a synced pattern, in which case the page only holds the reference.
+		if ( false === strpos( $post->post_content, 'dsgoSchema' )
+			&& false === strpos( $post->post_content, 'wp:block ' ) ) {
 			return;
 		}
 
--- a/includes/features/class-schema-output.php
+++ b/includes/features/class-schema-output.php
@@ -118,12 +118,31 @@ class SchemaOutput {
 			return array();
 		}
 
-		if ( false === strpos( $reference->post_content, 'dsgoSchema' )
-			&& false === strpos( $reference->post_content, 'wp:block ' ) ) {
+		if ( ! $this->may_contain_schema( $reference->post_content ) ) {
 			return array();
 		}
 
 		return $this->collect( parse_blocks( $reference->post_content ), $title, $seen );
+	}
+
+
+	/**
+	 * Could this content hold an opted-in block?
+	 *
+	 * A cheap string test used to skip the comparatively expensive
+	 * parse_blocks() on the overwhelming majority of posts. An opted-in block
+	 * carries the attribute name in its block comment — unless it lives in a
+	 * synced pattern, in which case only the reference is present here and the
+	 * attribute is in the referenced wp_block post.
+	 *
+	 * @param string $content Post content.
+	 * @return bool Whether the content is worth parsing.
+	 */
+	private function may_contain_schema( $content ) {
+		$content = (string) $content;
+
+		return false !== strpos( $content, 'dsgoSchema' )
+			|| false !== strpos( $content, 'wp:block ' );
 	}
 
 	/**
@@ -140,11 +159,7 @@ class SchemaOutput {
 			return;
 		}
 
-		// Cheap bail before the comparatively expensive parse. An opted-in block
-		// carries the attribute name in its block comment — unless it lives in
-		// a synced pattern, in which case the page only holds the reference.
-		if ( false === strpos( $post->post_content, 'dsgoSchema' )
-			&& false === strpos( $post->post_content, 'wp:block ' ) ) {
+		if ( ! $this->may_contain_schema( $post->post_content ) ) {
 			return;
 		}
 
--- a/includes/features/class-schema-output.php
+++ b/includes/features/class-schema-output.php
@@ -114,7 +114,10 @@ class SchemaOutput {
 			return array();
 		}
 
-		if ( 'publish' !== $reference->post_status ) {
+		// Mirrors core's render_block_core_block(), which refuses a pattern that
+		// is unpublished OR password-protected. Emitting schema for content core
+		// declines to render would disclose it through the back door.
+		if ( 'publish' !== $reference->post_status || ! empty( $reference->post_password ) ) {
 			return array();
 		}
 
@@ -156,6 +159,16 @@ class SchemaOutput {
 		$post = get_post();
 
 		if ( ! $post instanceof \WP_Post || ! has_blocks( $post->post_content ) ) {
+			return;
+		}
+
+		// A protected post is still singular, and get_post() still hands back
+		// the full content — WordPress only substitutes the password form at
+		// the_content, which runs long after wp_head. Without this the
+		// questions, answers and title would sit in view-source for anyone,
+		// password or not. Matches the guard in class-style-binding.php,
+		// class-query-bindings-helpers.php and the llms-txt classes.
+		if ( post_password_required( $post ) ) {
 			return;
 		}
 
--- a/includes/features/class-schema-output.php
+++ b/includes/features/class-schema-output.php
@@ -22,14 +22,37 @@ require_once DESIGNSETGO_PATH . 'includes/features/schema-builders.php';
 class SchemaOutput {
 
 	/**
-	 * Map a dsgoSchema value to its builder function.
+	 * Which builder handles which schema type, per block.
 	 *
-	 * @var array<string, string>
+	 * Keyed by block name first, deliberately. The allowlist otherwise exists
+	 * only in the editor UI and in the server-side attribute registration, and
+	 * neither constrains what parse_blocks() hands this collector: a
+	 * hand-written block comment (the editor's Code view suffices — no
+	 * unfiltered_html required) could put dsgoSchema on any block and have a
+	 * builder run against its children. Keying on the block name is what makes
+	 * "only blocks with a builder behind them" true at runtime rather than
+	 * merely in the UI.
+	 *
+	 * Keep the block names in step with includes/extension-configs/schema.php;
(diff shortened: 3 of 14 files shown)


jnealey-godaddy:
All three confirmed and fixed in `16ddb108`.

## Medium — `render_block()` during `wp_head`

Confirmed, and it is the real bug here. I reproduced it before changing anything: a probe block registered with a `render_callback` that counts its invocations, dropped into an accordion answer, ran **once during `wp_head`** and would run again in the loop.

```
$this->assertSame( 0, self::$renders ); // 1 before the fix, 0 after
```

Your reasoning about the second-order problem is right too — `render_block()` builds a fresh root `WP_Block` whose only context is `postId`/`postType` off the global, so a block relying on ancestor-supplied context renders wrong on top of running twice.

Fixed by serializing rather than rendering: `serialize_blocks()` reconstructs the stored markup and executes nothing. Block delimiters are HTML comments, which `strip_tags()` removes along with the tags, so the text comes out the same for the static content this is actually for. A dynamic block now contributes no text, which is the correct outcome — its output is not knowable at `wp_head` and is not part of the authored answer.

I preferred this to an allowlist of "safe" block types: an allowlist has to be maintained and fails open for any third-party block nobody thought about, whereas not executing anything is safe by construction.

`tests/phpunit/schema-no-render-test.php` pins it with the probe block, and a builder-level test asserts a dynamic sibling contributes nothing. Verified the emitted JSON-LD on a real page is unchanged by the switch.

## Low/Medium — `get_the_title()` breaks the purity contract

Confirmed, and your diagnosis of *how* it was visible is the part worth repeating: the test file itself demonstrated it. `test_howto_carries_the_post_title_as_its_name()` needed `go_to( get_permalink() )` while every FAQ test called the builder directly — the asymmetry was the tell.

The title is now a parameter, and `SchemaOutput::render()` passes `get_the_title( $post )` through `collect()`. Builders that do not need it ignore the extra argument.

Two new tests exercise it with no request state at all: one asserts the name comes from the argument, one asserts `name` is omitted rather than emitted empty when no title is given — claiming an empty `HowTo.name` would be worse than omitting a required property.

## Low — unenforced JS/PHP duplication

Confirmed. The comment asked a human to keep two lists in sync and nothing checked. The failure mode is exactly as you describe and is silent from the author's side: a control appears, they pick a type, the value is dropped because the attribute was never registered server-side.

`tests/phpunit/schema-config-parity-test.php` now parses `SCHEMA_TYPES` out of `constants.js` and asserts the block list matches `includes/extension-configs/schema.php`, plus that the attribute name/type/default agree and that every allowlisted block has a `block.json`. It reads the source file rather than a build artefact, since the assertion is about the file a developer edits.

Confirmed it catches the drift by adding `designsetgo/card` to the JS side only — the test fails with the intended message.

You are right that the wider extension pattern tolerates this risk elsewhere. I have deliberately not gone and retrofitted parity tests onto the other 18 configs in this PR; that is a separate sweep, and worth doing.

---

**Gate:** build, `lint:js`, `phpcs` (126 files), **16,164 unit tests**, **1,150 PHPUnit** — all green. Re-verified the rendered page output after the change: identical graph, one script tag.

jnealey-godaddy:
One confirmed and fixed in `2f2fd143`; two I checked and am pushing back on, with evidence. All three now have tests either way.

## Medium — synced patterns: confirmed, fixed

You're right, and this was the one that mattered. Reproduced before changing anything: an accordion in a `wp_block`, referenced from a page, emitted nothing at all.

`collect()` now expands `core/block` references. Because that opens a recursion hazard the original code didn't have, it carries a seen-set: a pattern referencing itself, or two patterns referencing each other, terminates instead of running until the request dies. Both are tested — the meaningful assertion in those two cases is that the test completes at all.

Also handled: nested patterns (a pattern inside a pattern), draft patterns, references to non-pattern post types, dangling refs, and `<!-- wp:block /-->` with no `ref`. Eight tests.

The cheap bail had to loosen — a page whose schema lives in a pattern has no `dsgoSchema` in its own content — so it now also passes when `wp:block ` is present. Still a `strpos` pair, still short-circuits the overwhelming majority of posts before `parse_blocks()`.

## Medium — breadcrumb depth: not a bug

I checked this before touching it, because if true it would be exactly the kind of accidental-correctness that bites later. It doesn't hold: **closing tags report the parent's depth, not the element's** — the breadcrumb stack is popped before the closer is visited. Dumping the walk for `<div><button><span class="t">Q</span><span class="i">LEAKED</span></button></div>`:

```
#tag:DIV            depth=3
#tag:BUTTON         depth=4
#tag:SPAN           depth=5   <- match, $found = 5
#text "Q"           depth=6   <- collected
#tag:SPAN(close)    depth=4   <- 4 < 5, breaks HERE
#tag:SPAN           depth=5   <- never reached
#text "LEAKED"      depth=6   <- never reached
```

So the loop terminates at the matched element's own closing tag, before the sibling is ever visited. The premise that "that sibling's breadcrumb count is identical to `$found`" is true of the sibling's *opening* tag, but the closer comes first and has already exited.

That said, your underlying worry — that this is only verified by markup that happens to have no text after the title — was fair, and it's now pinned rather than assumed. Three tests: a text-bearing sibling immediately after the title (`LEAKED`, asserted absent), deeply nested markup inside the title (collected in full), and an empty title element (yields `''`, not the sibling's text).

## Low — decode/strip ordering: deliberate, and swapping it regresses correctness

Agreed on the general WP guidance, and agreed it isn't exploitable here. But the swap has a cost that outweighs the benefit in this specific case, and I tested it rather than argue it. Applying your patch verbatim:

```
-'Use <b>bold</b> for emphasis.'
+'Use bold for emphasis.'
```

An answer whose saved markup is `Use &lt;b&gt;bold&lt;/b&gt;` **displays** `Use <b>bold</b>` on the page. Decoding first turns those entities into real tags, and `strip_tags()` then deletes them — so the schema text silently stops matching the visible content, which for structured data is the actual failure mode Google cares about. Stripping first, then decoding, keeps the extracted text equal to what the reader sees.

Rather than rely on the downstream net as an unstated assumption, I've made it explicit and tested. A question whose saved markup is `&lt;/script&gt;&lt;script&gt;alert(1)&lt;/script&gt;` now asserts three things: no real element boundary appears in the head, the JSON still parses, and the author's text survives verbatim inside the JSON string. The `</` → `<\/` rewrite covers every source of that sequence at once — decoded entities included — which is why it's the right place for the guarantee rather than the ordering.

Happy to revisit if you'd rather have the ordering match convention and accept the text loss, but I'd want that to be a deliberate trade rather than a silent one.

---

**Gate:** build, `lint:js`, `phpcs` (126 files), **16,164 unit tests**, **1,163 PHPUnit** — all green. Schema coverage is now 80 tests.

jnealey-godaddy:
Both confirmed and fixed in `c972887c`.

## Low — `HowTo.name` not entity-decoded

Confirmed, and the inconsistency is sharper than "minor" once you see it on one page. Dumping the emitted graph for a post titled `Tea &amp; Coffee &mdash; a guide` containing a question `Tea &amp; coffee?`:

```
howto_name => 'Tea &amp; Coffee &mdash; a guide'   <- not decoded
step_name  => 'Tea & coffee?'                      <- decoded
step_text  => 'Both & more.'                       <- decoded
```

So a single `HowTo` node carried both forms simultaneously. I'd call it a data-quality *bug* rather than an inconsistency: `&amp;` and `&mdash;` are not the characters the author wrote, and a consumer reading that name gets literal entity text — the schema stops matching the visible page, which is the thing structured data is judged on.

The title now gets the same treatment as everything else — `wp_strip_all_tags()`, then `html_entity_decode()`, then whitespace collapsed, which the title path was also missing.

Three tests: entities decoded (asserting the name and the step agree, since the agreement is the actual invariant), markup stripped from the title, and a title that reduces to empty omitting `name` rather than emitting `""`. Confirmed the first fails against the old code:

```
-'Tea & Coffee — a guide'
+'Tea &amp; Coffee &mdash; a guide'
```

## Low — duplicated cheap-bail heuristic

Confirmed. `render()` and `collect_from_reference()` each carried the same pair of `strpos()` calls, which is exactly the shape that drifts the next time a marker changes — and the failure mode would be quiet: patterns silently stop being scanned, or every post gets parsed.

Extracted to a single private `may_contain_schema( $content )`, used by both. Same short-circuit behaviour, one definition. The explanation of *why* both markers are needed now lives with the check rather than being duplicated in two comments.

---

**Gate:** build, `lint:js`, `phpcs` (126 files), **16,164 unit tests**, **1,166 PHPUnit** — all green. Schema coverage is 83 tests.

jnealey-godaddy:
Confirmed green in CI — and you've found a real defect, just not in the code. Fixed in `5871d347`.

## The test is green

`WordPress latest / PHP 8.3` on `c972887c`: **1167 tests, 0 failures**, which includes `test_title_extractor_stops_at_the_end_of_the_matched_element` with its literal `LEAKED` sibling. It is not skipped and not conditional.

## Why — the token order

The breadcrumb semantics do differ from what my docblock claimed. **WP_HTML_Processor pops the breadcrumb stack before visiting a closing tag**, so the matched element's own closer reports its PARENT's depth, and it is visited before the sibling ever opens. Traced token-by-token over your exact example:

```
#tag:DIV           depth=3
#tag:BUTTON        depth=4
#tag:SPAN          depth=5   <- match, $found = 5
#text "Q"          depth=6   <- collected
#tag:SPAN(close)   depth=4   <- 4 < 5, loop breaks HERE
#tag:SPAN          depth=5   <- never reached
#text "LEAKED"     depth=6   <- never reached
```

Your reading — that a following sibling opens at the same depth as the match — is correct. What it misses is that the closer arrives first and has already exited. So the loop does stop at the title's own closing tag; it just gets there via the pop rather than via a tag-name comparison.

## The actual defect was mine

You put your finger on something I should have caught: **the docblock and the test asserted opposite things.** I wrote that docblock from the first round's framing and only verified the mechanism afterwards, then never went back and corrected the prose. A comment that contradicts its own assertion is worse than no comment — it sent you down the same path twice, and it would have done the same to the next reader.

Both the test docblock and the code comment now carry the trace above and say explicitly why the comparison terminates. That is the real fix here.

## One extra guard

I've added `test_title_extractor_survives_a_nested_element_of_the_same_tag`, because the natural remedy for the bug you suspected — "break on a closing `<span>`" — would introduce one: the inner span's closer arrives first, so keying off tag name alone truncates a title containing nested markup. Any future rewrite of the stop condition now has to keep that passing.

The extractor is therefore pinned by five tests: a text-bearing sibling (asserted absent), nested markup collected in full, nested SAME-tag markup, an empty element, and a missing element. The real markup's icon is `<svg>`-only and could not catch a regression, which is exactly why these are synthetic.

I did consider restructuring the loop to key off the matched element's closer explicitly rather than depth. I decided against it: it would need manual open/close counting with void-element handling, which adds failure modes to code that is verified correct and now guarded — and the `LEAKED` test fails loudly if the breadcrumb behaviour ever changes underneath us, which is the protection that actually matters.

---

**Gate:** `phpcs` (126 files), **1,167 PHPUnit** — green.

jnealey-godaddy:
Confirmed and fixed in `e46d6112`. Good catch — this one was a genuine disclosure and I should have caught it myself, given the codebase already guards it in five places.

## The leak

Reproduced before changing anything. A post with `post_password` set, containing an opted-in accordion, emitted the full graph into `wp_head`:

```
Failed asserting that '…' does not contain "application/ld+json".
```

Your explanation of why is exactly right: `is_singular()` is still true, `get_post()` still returns the full object, and WordPress only substitutes the password form at `the_content` — which runs long after `wp_head`. So the questions, answers and title were readable in view-source by anyone, password or not.

Fixed with the bail you suggested, alongside the existing `is_singular()` / `WP_Post` checks, matching `class-style-binding.php:146`, `class-query-bindings-helpers.php:94` and the two llms-txt classes.

## A second instance, same class

Following the thread turned up one the review did not mention. `collect_from_reference()` — the synced-pattern expansion added earlier in this PR — checked a pattern's `post_status` but not its password. Core's `render_block_core_block()` refuses both:

```php
// wp-includes/blocks/block.php:42
if ( 'publish' !== $reusable_block->post_status || ! empty( $reusable_block->post_password ) ) {
```

So a password-protected pattern embedded in a public page would have had its questions and answers emitted into the head, even though core declines to render it on the page. Now mirrors core exactly.

## Tests

Four, in the two relevant suites:

- no output for a password-protected post
- no output for a password-protected HowTo (asserting on the JSON-LD payload and step content — the post title legitimately appears in `<title>`, so asserting its absence outright would have been a bogus test)
- no output for a password-protected synced pattern
- **output restored once the password cookie is present** — this one matters as much as the others, since a guard that simply never emits would pass all three negative tests

That last one needs the cookie to hold a *phpass* hash specifically: `post_password_required()` rejects anything not starting with `$P$B` before it checks the value, so `wp_hash_password()` on modern WP returns bcrypt and would have read as "still locked" — making the test pass for the wrong reason.

---

**Gate:** `phpcs` (126 files), **1,171 PHPUnit** — green. Schema coverage is 88 tests.

jnealey-godaddy:
Both confirmed and fixed in `6641df09`.

## Medium — dispatch ignored `blockName`

Confirmed, and it reproduces with real output rather than just in theory. A hand-written `core/group` carrying the attribute, wrapping genuine accordion items, emitted a full `FAQPage`:

```html
<!-- wp:group {"dsgoSchema":"faq"} -->
  <!-- wp:designsetgo/accordion-item -->…<!-- /wp:designsetgo/accordion-item -->
<!-- /wp:group -->
```

Your framing of the impact is right — no escalation, no XSS, output still neutralised — but it broke the design principle this PR argued for in its own description, which is the part that matters. I had enforced "only blocks with a builder" in the editor UI and in the attribute registration, and neither of those constrains what `parse_blocks()` hands the collector at runtime. The one place it needed to hold, it didn't.

`BUILDERS` is now keyed by block name first:

```php
private const BUILDERS = array(
    'designsetgo/accordion' => array(
        'faq'   => 'designsetgo_schema_build_faq',
        'howto' => 'designsetgo_schema_build_howto',
    ),
);
```

so the allowlist is the dispatch table rather than a parallel claim about it.

Two tests: the smuggled `core/group` emits nothing, and the real accordion still emits — the second matters, since keying on the wrong thing would pass the first while breaking the feature. Confirmed the negative test fails if the block-name check is bypassed.

That list is now the third copy of the allowlist, so `Schema_Config_Parity_Test` covers it too. All three — JS constants, PHP attribute config, runtime builder map — are asserted equal, via a new public `SchemaOutput::supported_blocks()`.

## Low — duplicated normalization

Agreed, and your reasoning for it is the reason I took it: this exact logic had already been fixed in one place and missed in the other, which is how the same page ended up emitting `Tea &amp; Coffee` beside `Tea & coffee?`.

Extracted to `designsetgo_schema_normalize_text()`, now the single definition. The docblock records why the step order is what it is — separate at tag boundaries first (or `<p>One.</p><p>Two.</p>` fuses to `One.Two.`), strip, and only then decode, since decoding first would let `strip_tags()` delete text the page actually displays.

Both call sites delegate, so a future change to that sequence can only be made once.

---

**Gate:** build, `lint:js`, `phpcs` (126 files), **1,174 PHPUnit** — green. Schema coverage is 91 tests.

jnealey-godaddy:
Both are decisions rather than defects, so here they are — made explicitly, measured where measurable, and recorded in the code rather than only in this thread. `df330de2`.

## Scope: templates stay out, deliberately

You're right that it's the same silent-no-op shape as the synced-pattern bug. But I don't think the same answer follows, and it's worth saying why rather than just extending the walk.

A block in a template or template part appears on **every page using that template**. `FAQPage` markup is supposed to describe the page it sits on, so emitting one site-wide FAQ across every URL is precisely the structured-data spam this feature's opt-in default exists to prevent — the same reasoning in the PR description for why nothing is emitted unless an author deliberately chooses a type. Collecting from templates would make that easy to do by accident and hard to notice.

The synced-pattern case is different in kind: a pattern is placed *into a specific page's content*, so its schema is about that page. That's why expanding references was right and expanding templates isn't.

Sites that genuinely want template-level schema have `designsetgo_schema_nodes`, where the decision is explicit and per-site rather than implicit and global.

Recorded as a `SCOPE:` note in the class docblock so the next reader finds the reasoning, not just the absence.

## Cost: measured, and not worth a cache

Rather than agree in principle, I benchmarked `render()`:

| scenario | per request |
|---|---|
| page with **no schema** (the overwhelming common case) | **0.0008 ms** |
| 3 synced patterns × 20 Q&A pairs = 60 questions, + 100 other blocks | **7.6 ms** |

The `strpos()` pair does its job — under a microsecond for a page that isn't participating. The heavy figure is a deliberately unrealistic page; a normal FAQ is a fraction of it.

Against that, a transient has to be invalidated when the post changes **and** when any pattern it references changes — which needs a reverse index of which posts reference which patterns. That's a real mechanism with a real failure mode, and stale structured data is worse than the 7.6ms it saves: wrong answers served to crawlers, invisible to the author. I'd rather not buy that complexity at this price.

Both figures are in the docblock as a `COST:` note, so if adoption does make this hot the next person starts from a number instead of re-deriving one.

Happy to revisit either if you disagree — particularly the template one, since it's a product judgement as much as a technical one.

---

**Gate:** `phpcs` (126 files), **1,174 PHPUnit** — green.

PR #571, opened 2026-09-18
fix(abilities): close the save() drift class, and split the 6.9k-line inserter

jnealey-godaddy:
## Description

PR #565 fixed eleven separate cases of one bug: `Block_Inserter` re-implements each block's `save()` in PHP, and wherever the mirror drifted, generated markup failed validation and the author saw "Attempt Recovery".

This addresses the *class*. Three places in the abilities layer restate by hand something that already has a machine-readable source of truth, with nothing asserting the restatement is faithful:

| Restatement | Source of truth |
|---|---|
| `Block_Inserter`'s PHP mirror of every `save()` | `src/blocks/{slug}/save.js` |
| `Block_Configurator`'s rich-text allow-lists | `source: "html"\|"text"` in `block.json` |
| `Configure_Shape_Divider::VALID_SHAPES` | `src/blocks/section/utils/shape-dividers.js` |

Each now has a CI guard that diffs it against its source. Then the 6,902-line file the first one lived in is split.

## Type of Change

- [x] Refactor

## What the attribute matrix found

The existing fixture covered each block at its **default** attribute values — 144 payloads. The matrix flips **every attribute in the registry, one at a time**, and round-trips it through the same PHP → fixture → JS `isValid` loop: **4,073 payloads**.

It opened at **375 failures across 46 blocks**. All were live drift in shipped code:

- **5 universal extensions the mirror reproduced not at all** — responsive hide, custom CSS, reveal-on-hover, background video, clickable-group; max-width only on 3 of ~70 blocks. Setting "hide on mobile" through an ability wrote markup the editor rejects.
- **12 blocks declare `supports.align` and never called `align_class()`**.
- **`extraProps` vs block props** — Modal's custom class, extension classes and alignment landed on its content div instead of the root.
- **Preset shorthand reaching markup raw.** Not just invalid blocks: `var:preset|…` is unparseable CSS, so those colours never rendered on the front end either.
- **`counter-group::gap`** — `intval() . 'px'` reproduced the `"32px"` default exactly, and turned `2rem` into `2px`. A wrong implementation that agrees with the default is invisible to per-block coverage; this is the case the matrix exists for.
- **`counter-group::alignContent`** read an `alignment` attribute the block has never declared.
- **`table-of-contents`** wrote no style attribute at all.
- **`card`** rendered no image element at all.
- **`icon-list`** omitted `data-dsgo-icon-style`, so markup matched an **older deprecation** and migrated silently on open. The block was *valid* — no validity check could have caught it. It surfaced only because the suite fails on an unexpected `console.info`, which is what WordPress logs when a deprecation migrates a block.

…plus a long tail across form-builder, modal, progress-bar, icon-button, modal-trigger, row, slide, image-accordion and blobs.

**Three faults were in the harness or the probes, not the product, and were fixed before the baseline was recorded** — a matrix that reports drift the product does not have is worse than no matrix:

1. `tools/regenerate-patterns` imports only `block-animations`, deliberately. Without the other 16 save-affecting extensions, `save()` in the test could not emit their props while PHP did — 138 phantom failures.
2. `textColor` / `backgroundColor` / `borderColor` / `fontSize` are core's preset-**slug** attributes; the `var:preset|…` shorthand belongs in `style.color.*` — 102 phantom failures.
3. The probe table's own format was ambiguous for array-valued attributes and probed every one of them with the wrong shape.

## Rich-text fidelity

Block validation is **structurally blind** to this class. For a `source: "html"` attribute the value lives in the markup, so stripping tags before writing means the parser reads it back stripped, `save()` reproduces exactly that, and the block is perfectly valid — with the author's formatting gone. `Care <em>begins</em>` shipped as `Care begins`.

Six attributes were classified wrong and now keep their markup: card `title`/`subtitle`/`bodyText`, timeline-item `date`/`title`, accordion-item `title`.

accordion-item is the instructive one: the sanitizer stripped `<em>` **and** the inserter emitted with `esc_html`. The two bugs cancelled out, which is why neither was visible — hence the test checks sanitization and serialization separately.

## The split

`class-block-inserter.php`: **6,902 → 2,689 lines**. A 3,111-line `switch` becomes 50 serializer classes behind `Serializer_Registry`.

The registry's value is that the set of serializable blocks is now **enumerable**. `get_serialization_gap()` used to answer "can this block be inserted?" by running the switch and checking for `null`.

**Both fixtures regenerate byte-identical through the move.** That is why the matrix was built before the refactor rather than after — against the old 144-entry fixture the same move would have proven a fraction as much.

## What this does NOT do

Two real faults are recorded and deliberately left, because both change stored markup and need deprecations:

- `counter`'s `save()` writes a preset shorthand into a CSS custom property, so that colour never applies on the front end. The mirror now reproduces it faithfully, bug included — the mirror's contract is fidelity to `save()`, not being more correct than it.
- `counter`'s `showIcon` and `advanced-heading`'s `animatedHeadline` inline SVG from a JavaScript icon library with no PHP equivalent. Both are refused at insert time rather than approximated, as `pathType: custom` already was.

The matrix also proves one attribute at a time, not combinations. Attributes that only affect `save()` alongside another round-trip without exercising anything; that stays the job of the hand-authored scenarios.

## Testing

- `vendor/bin/phpunit` — **1,695 passed**, 6,710 assertions, 1 documented skip
- `npm run test:unit` — **3,678 passed**, 172 suites
- `composer lint` (PHPCS) — clean
- `composer analyse` (PHPStan) — clean

New guards, each a two-way key diff so it cannot rot:

- `abilities-attribute-probe-table-test.php` — every registry attribute has a probe or a reasoned skip
- `abilities-attribute-matrix-fixture-test.php` + `ability-attribute-matrix.test.js` — the 4,073-payload round-trip
- `abilities-rich-text-policy-test.php` / `abilities-rich-text-fidelity-test.php`
- `abilities-shape-enum-test.php` — parses the JS directly, so a broken generator cannot pass by agreeing with itself
- `abilities-serializer-registry-test.php`
- `block-inserter-universal-extension-props-test.php`

Both fidelity and enum tests were verified to actually fail when the bug is reintroduced.

- [ ] Tested on frontend
- [ ] Tested responsive behavior
- [x] No console errors
- [x] No PHP errors

The boxes above are unchecked because no manual sweep was run; the case for this branch is the automated round-trip, and the fixtures assert byte-level equivalence against the real `save()`.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have followed the patterns in CLAUDE.md
- [x] Accessibility: WCAG 2.1 AA compliant
- [x] Security: All user input is validated and sanitized
- [x] Internationalization: All user-facing strings use `__()`

## Additional Notes

Rebased onto main after #565 merged. Its only conflict was `section::boxWidth` / `contentPosition`, which main added to the switch this branch deletes; both were ported into `Section_Serializer`.

That rebase also exercised the new guard for real: main's `card::imageFallbackAlt` failed the probe-table key diff immediately, exactly as intended for a new attribute arriving from upstream.

Design and implementation notes are in `docs/plans/2026-09-17-abilities-drift-prevention.md` and its `-implementation` companion.


jnealey-godaddy:
@claude do a pr review

PR #575, opened 2026-09-18
fix(ci): restore the PHP 7.4 and 8.0 test jobs (Debian bullseye archived)

jnealey-godaddy:
Every PHP 7.4 and 8.0 job in the WordPress test matrix has failed since about Sep 14, before any test runs. Main's last fully green CI run was the 2.7.3 release.

**Cause.** The `wordpress:php7.4` and `wordpress:php8.0` images are built on Debian 11 (bullseye). Its LTS ended on 2026-08-31, and the `deb.debian.org` security pool now returns 404 for packages the images still list, so wp-env's `apt-get install $PHPIZE_DEPS` step exits 100.

**Fix.** wp-env 11.x already fixes this upstream (`lib/runtime/docker/docker-config.js`). This backports the same three Dockerfile lines to the wp-env 10.x generator we pin, using a postinstall script next to the existing composer-audit patch:
- rewrite bullseye to `archive.debian.org`
- drop the `bullseye-security` and `bullseye-updates` suites, which aren't on the archive yet

The script is idempotent, and it becomes a no-op once wp-env is upgraded to 11.x, when it can be deleted.

**Verified**
- On the real `wordpress:php7.4` image, the patched sources make `apt-get install $PHPIZE_DEPS` exit 0.
- The script inserts the block after the buster lines in wp-env 10.35's `init-config.js`. A second run changes nothing.
- The PHP 7.4 / 8.0 jobs on main after merge are the end-to-end proof. This PR's own run only covers WordPress latest on PHP 8.3.


jnealey-godaddy:
Full matrix verified before merge: I pushed this branch to `release/ci-bullseye-check` (CI runs the full matrix on `release/**`), and run [35380337023](https://github.com/jnealey-godaddy/designsetgo/actions/runs/35380337023) passed all 20 WordPress × PHP jobs, including every PHP 7.4 and 8.0 job. That temporary branch has been deleted.

PR #576, opened 2026-09-18
fix(blocks): keep existing overlays at their original strength

jnealey-godaddy:
#565's overlay change (`c9e61464`, unreleased) lowered the default container overlay from 0.8 to 0.65, and to 1 for colours that carry their own alpha. The intent was new blocks, but existing sites would have changed in two ways:

1. **On update, with no edits.** An overlay that stores no inline opacity, such as a Style Kit `is-style-overlay-*` variation that doesn't set its own, fell back to the stylesheet default, which had moved to 0.65. Every such page would lighten the moment the plugin updated.
2. **On the next edit.** The deprecations reproduced the stored `0.8` markup and passed attributes straight through, so the next save wrote the lighter default.

**Change**
- Section, Row, Grid and Scroll Accordion Item get an integer `overlayOpacity` percentage. When it's set, it wins over the colour-aware default. New blocks leave it unset. This is the same name and 0–100 format that Scroll Slides, Image Accordion and Slide already use.
- `withLegacyOverlayOpacity()` wraps every deprecation entry that reproduces pre-2.8.0 markup, so old content migrates to `80`. Deprecations don't cascade, so this covers all of them, not just the newest. Section `v10` is excluded because it reuses the current `save()`, so content it claims already has the new opacity.
- The stylesheet fallback goes back to `0.8` in `style.scss` and `editor.scss` for all four blocks.
- `Serializer_Support::overlay_opacity()` is the PHP twin for the Abilities inserter. It's integer-only on both sides, because `n/100` prints the same in JS and PHP for every integer, but `33.3` doesn't (`0.33299999999999996` vs `0.333`).

**Verified**
- Jest: 3,776 tests pass (27 new). The deprecation suite now asserts that stored `0.8` content migrates to `overlayOpacity: 80` and re-saves byte-identical. It also asserts that new blocks stay unset, and that Section `v10` content isn't pinned.
- PHPUnit: the overlay suite passes with shared `percentCases`, and the Abilities fixtures were regenerated. Only the four new `overlayOpacity` matrix entries are kept; the ~50 `anchor`/`style` entries that WP 7.1 adds are version-specific, and CI generates on older WordPress. phpcs and PHPStan are clean.
- In a browser on WP 7.1 with real stored 2.7.x markup:
  - Section, Row and Grid legacy overlays paint at 0.8 after the update.
  - A class-only overlay without its own opacity falls back to 0.8.
  - A new Section paints at 0.65.
  - Editing and saving the page: no recovery prompt, the three legacy blocks are stored with `"overlayOpacity":80` and `0.8`, and the new block keeps `0.65`.
  - After a reload the content round-trips exactly, with no deprecation step.


jnealey-godaddy:
Full matrix verified: combined with #575's CI fix on a temporary `release/overlay-preserve-check` branch, run [35381703655](https://github.com/jnealey-godaddy/designsetgo/actions/runs/35381703655) passed all 30 jobs, including all 20 WordPress × PHP 7.4–8.3 test jobs. The temporary branch has been deleted.

